PR #3, opened 2021-05-23
Adds Workflows

Shivansh-007:
Adds workflows which would:

- docker build and push to GHCR.io
- linting which would lint all pull requests and pushes to main

**MERGE #2 before this**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,92 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+# This will cancel any previous running actions.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  lint:
+    name: Run pre-commit & flake8
+    runs-on: ubuntu-latest
+    env:
+      # Configure pip to cache dependencies and do a user install
+      PIP_NO_CACHE_DIR: false
+      PIP_USER: 1
+
+      # Disable Poetry virtualenv creation
+      POETRY_VIRTUALENVS_CREATE: false
+
+      # Specify explicit paths for python dependencies and the pre-commit
+      # environment so we know which directories to cache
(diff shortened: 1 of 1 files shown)


bast0006:
> .pre-commit-config.yaml is not a file

We should probably have a pre-commit config file set up before we accept this, otherwise no idea if this works. Checks should fail though since we're not flake8 supporting.

Shivansh-007:
> > .pre-commit-config.yaml is not a file
> 
> We should probably have a pre-commit config file set up before we accept this, otherwise no idea if this works. Checks should fail though since we're not flake8 supporting.

Already done, [here](https://github.com/onerandomusername/modmail/blob/main/.pre-commit-config.yaml), I think it is failing because this branch is not upto date with `main` and hence doesn't have that file.

bast0006:
I'm not able to merge in those changes but you should be able to, IIRC there's a new pretty button for it too near the top (update branch or something)

onerandomusername:
> 
> 
> I'm not able to merge in those changes but you should be able to, IIRC there's a new pretty button for it too near the top (update branch or something)

There's not a button for this unless its a fork.

bast0006:
>        # Check out the current repository in the `sir-lancebot` subdirectory
 These also need stripping/editing/licensing

.github/workflows/build.yaml
```diff
@@ -0,0 +1,59 @@
...
+          password: ${{ secrets.GITHUB_TOKEN  }}
+
+      # Build and push the container to the GitHub Container
+      # Repository. The container will be tagged as "latest"
+      # and with the short SHA of the commit.
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: modmail/
+          file: sir-lancebot/Dockerfile
```
onerandomusername, line 51: This is for the wrong repo

onerandomusername (changes requested):
I don't see a reason to have publishing to docker at the moment.

I think publishing to docker should be added on a later time, after a DockerFile is added.

bast0006 (approved):

onerandomusername (approved):
LGTM

PR #7, opened 2021-05-25
Merge Site repo into modmail

### LINKED ISSUE #6: Merge discord-modmail/site into modmail
After lots of discussion, we've decided that the best way to manage this project is to merge the site into the bot.

This comes with multiple benefits, as we don't have to maintain database schemas on both repos, have a seperate backend for each, and makes it easier to deploy. 

As of now, it is our goal to make the site and bot mostly independent from one another, and running the site shall be optional.

bast0006:
Closes #6


bast0006:
This needs post merge cleanup pretty badly

onerandomusername (changes requested):
This looks pretty good. 

I think that the site stuff should all be moved to a `/site` folder. If that name causes conflicts, `/website` would be another possible name.

One possibilty to launch both the bot and the site would be to use starlette's on_start methods, but I think we need to have a strategy for being able to run just the bot or just the site.

onerandomusername:
I left out of my review, but pyproject and poetry lock need to be updated.

Shivansh-007 (approved):

PR #9, opened 2021-05-26
Add shared startup functionality

bast0006:
to run bot and site with one command and in the same event loop


dawnofmidnight (approved):
It works

PR #11, opened 2021-05-31
Adds Discord OAuth2 login to site

Shivansh-007:


.env.example
```diff
@@ -1,5 +1,13 @@
...
 MODMAIL_GUILD_ID=1234567890
-OWNERS='hello world'
+
+# DATABASE
 CONNECTION_URI="mongodb+srv://mongodburi"
+
+# DISCORD LOGIN
+CLIENT_ID = "YOUR APP'S CLIENT ID HERE"
+CLIENT_SECRET = "YOUR APP'S CLIENT SECRET HERE"
+REDIRECT_URI = "YOUR DISD LOGIN REDIRECT URI"
```
bast0006, line 13: This should be `"https://yourdomain.com/callback"` so that /callback is not forgotten. That or we should construct it down where we use it, so you just put "yourdomain.com" as an example.

modmail/bot.py
```diff
@@ -25,7 +26,7 @@ def __init__(self, **kwargs):
         super().__init__(command_prefix=self.get_prefix, **kwargs)
         self.http_session = ClientSession()
 
-    async def get_prefix(self, message: discord.Message = None) -> str:
+    async def get_prefix(self, message: discord.Message = None) -> t.List[str]:
```
bast0006, line 29: TBH this should be in a separate PR.

website/constants.py
```diff
@@ -0,0 +1,13 @@
+from os import environ
+from typing import NamedTuple
+
+import dotenv
+
+dotenv.load_dotenv()
+
+REDIRECT_URI = environ.get("REDIRECT_URI")
```
bast0006, line 8: Is there a reason we're separating out the client namedtuple like this? And not constructing an instance? What purpose does this serve except for categorization? I'm not sure what benefit pydantic would have here, but I don't think we should be using Client as a class just for namespacing. CLIENT_ID and CLIENT_SECRET are fine constants

bast0006 (dismissed):

modmail/bot.py
bast0006: Not sure if it's required for lints to pass though. If it is, someone ought push this to main and merge that in so lints pass.

Shivansh-007: > TBH this should be in a separate PR.

I thought it won't cause much different keeping this small change, I think lets keep it in this PR for the moment and I can remember that for future.

> Not sure if it's required for lints to pass though. If it is, someone ought push this to main and merge that in so lints pass.

No, lint won't fail on this as the linters don't check whether the annotations are correct or not.

website/constants.py
Shivansh-007: > Is there a reason we're separating out the client namedtuple like this? And not constructing an instance? What purpose does this serve except for categorization? I'm not sure what benefit pydantic would have here, but I don't think we should be using Client as a class just for namespacing. CLIENT_ID and CLIENT_SECRET are fine constants

I kept the "client" side constants inside the namedtuple to keep them as a group.

bast0006: Well, I think it's not worth the effort at this time to group them. However it doesn't matter that much. But they shouldn't be a NamedTuple. We're never constructing it, so it should just be a bare class to reduce imports and potential confusion. But it'd be better without grouping

modmail/bot.py
onerandomusername: In the future, this kind of edit should be in a separate pr.

Even more so, given that this is a site pr and not a bot pr.

Its okay to merge in this case.

pyproject.toml
```diff
@@ -23,14 +23,15 @@ python = "^3.8"
 aiohttp = { extras = ["speedups"], version = "^3.7.4" }
 colorama = "^0.4.3"
 coloredlogs = "^15.0"
-"discord.py" = "1.6.0"
+"discord.py" = "1.7.0"
```
onerandomusername, line 26: Dependency updates should be in their own pull requests in the future.

website/main.py
```diff
@@ -1,10 +1,14 @@
...
+from starlette.responses import JSONResponse, RedirectResponse
 from starlette.routing import Mount, Request, Route
 from starlette.staticfiles import StaticFiles
 from starlette.templating import Jinja2Templates
+from starlette_discord.client import DiscordOAuthClient
 
+from website.constants import CLIENT_ID, CLIENT_SECRET, REDIRECT_URI
 
 templates = Jinja2Templates(directory="website/templates")
+client = DiscordOAuthClient(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI)
```
onerandomusername, line 11: I'm not sure that `client` is the best variable name here. Perhaps something like `discord_client`?

onerandomusername (changes requested):
Everything looks okay to me.

In the future, I think that dependency updates and bot changes should be separate from site changes. 

Updating discord.py to v1.7 and changing the typing in the bot should be done separately from site prs.

onerandomusername (approved):

bast0006 (approved):

PR #13, opened 2021-06-01
Add Docker Setup

Shivansh-007:


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM tiangolo/uvicorn-gunicorn-starlette:python3.8
+
+# Install Poetry
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_HOME=/opt/poetry python && \
```
onerandomusername, line 4: The new installer for poetry is located at `install-poetry.py`

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+### Docker compose setup file
```
onerandomusername, line 1: Ideally, this file should be made for production, code autoreloading can be in an override file added by the user, and documented in `CONTRIBUTING.md`

.env.example
```diff
@@ -11,3 +11,6 @@ CONNECTION_URI="mongodb+srv://mongodburi"
 CLIENT_ID = "YOUR APP'S CLIENT ID HERE"
 CLIENT_SECRET = "YOUR APP'S CLIENT SECRET HERE"
 REDIRECT_URI = "https://yourdomain.com/callback"
+
+# OLLI
+WEBHOOK_URL = "YOUR WEBHOOK URL HERE"
```
onerandomusername, line 16: No longer using olli, this indicator can be removed from here

onerandomusername (changes requested):
Looks pretty good to me. A few changes, and its ready for merging.

onerandomusername:
Blocked by #32, since this adds features to the website.

Shivansh-007:
Sooo should I remove them from this PR or will you merge this and remove them from that PR? I had go with the second option 😉

docker-compose.yml
Shivansh-007: NOt necessarily, it can be used for development also, it gives much easier support when one only wants to develop with the bot, looking into the current situation, not after #32.

I also didn't make a contributing guide since there is none right now, once someone makes it, I will add the docker section to it.

dawnofmidnight, line 1: Isn't it implied exactly what this file is?

onerandomusername: Line #3 and line #4.

A production docker file shouldn't have code autoreloading.

dawnofmidnight: I'm not referring to the actual content of the file, just this specific comment. The name of the file implies that it's for docker compose.

onerandomusername: Ah, I didn't realise this wasn't replying to my review, sorry.

onerandomusername: Line 3 and line 4.

A production docker file shouldn't have code autoreloading. That's my biggest concern here.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM tiangolo/uvicorn-gunicorn-starlette:python3.8
```
onerandomusername, line 1: We aren't using gunicorn,  uvicorn,  or starlette anymore.

Perhaps from alpine or some other lightweight image.

.gitignore
```diff
@@ -133,5 +133,11 @@ dmypy.json
 # logs
 logs
 
+# Editors
+.vscode
```
onerandomusername, line 137: This was changed in main and is now duplicated in this file.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Install Poetry
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | POETRY_HOME=/opt/poetry python && \
+    cd /usr/local/bin && \
+    ln -s /opt/poetry/bin/poetry && \
+    poetry config virtualenvs.create false
+
+# Copy using poetry.lock* in case it doesn't exist yet
+COPY ./pyproject.toml ./poetry.lock* /app/
```
onerandomusername, line 10: As the current code,  we also want to copy `./config.toml` (if it exists) which contains the bot configuration.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: "python -m modmail"
+    volumes:
+      - .:/app
+    tty: true
+
+volumes:
+  staticfiles:
```
onerandomusername, line 20: There's nothing here,  so we could probably remove it.

onerandomusername (changes requested):

Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.9-slim
+
+# Set pip to have cleaner logs and no saved cache
+ENV PIP_NO_CACHE_DIR=false
+    # POETRY_VIRTUALENVS_CREATE=false
```
Shivansh-007, line 5: Why is this commented out?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# Install project dependencies
+
+# Copy dependencies and lockfile
+COPY pyproject.toml poetry.lock ./
+RUN poetry install --no-dev --no-interaction --no-ansi --no-root
+
+# Copy the source code in last to optimize rebuilding the image
+COPY . .
+
+ENTRYPOINT ["poetry"]
```
Shivansh-007, line 23: I have never done this on Dockerfiles in my past projects, don't see how this is needed here, actually, I don't see why this is not needed here too 😅. Mind explaining?

onerandomusername: Due to line 5 causing issues, the entry point on my machine had to be into the venv. This shouldn't be a problem, but it somehow is. I think it may be because some of our dependencies do not agree with poetry.

Shivansh-007: Uhhh okay, am still not clear. I will look around the docs and ask joe, if I understand the issue in a better way.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# Create the working directory
+WORKDIR /modmail
+
+# Copy the source code in last to optimize rebuilding the image
+COPY . .
+
+# Install project dependencies
+RUN poetry install --no-dev
+
+CMD ["python", "-m", "modmail"]
```
dawnofmidnight, line 29: Shouldn't it be `poetry run python -m modmail`?

Shivansh-007: We don't make poetry virtualenvs, on line 5, by setting `POETRY_VIRTUALENVS_CREATE` to `false`. Basically, what poetry does is, if this is set to `false` it uses `pip` to install the deps and if this is set to `true` (default) then it would create the virtualenvs in your cache (those `.venv` folders).

dawnofmidnight: Ah, ok

onerandomusername (approved):
LGTM

PR #27, opened 2021-07-29
update export task to not infintely loop

onerandomusername:
If `task export` was ran with `poetry run`, it was possible to error poetry in a way that never ended.


Shivansh-007 (changes requested):
It is really hard to go through the tasks, can we break them into sections with the help of comments and indent them multiline, this format:

```toml
[tool.taskipy.tasks]
export = { 
    cmd = "echo > requirements.txt && poetry export --without-hashes >> requirements.txt", 
    help = "Export installed packages in requirements.txt format" 
}
```

onerandomusername:
...

onerandomusername:
> It is really hard to go through the tasks, can we break them into sections with the help of comments and indent them multiline, this format:
> 
> ```toml
> [tool.taskipy.tasks]
> export = { 
>     cmd = "echo > requirements.txt && poetry export --without-hashes >> requirements.txt", 
>     help = "Export installed packages in requirements.txt format" 
> }
> ```

When I tried this I was getting errors, that it was not available pyproject.toml file.

Theres another format, if I recall correctly,  which is `[tool.taskipy.task.export]` but that ends up being so verbose.

Shivansh-007:
Umm alright, then just break them into sections with comment headers.

onerandomusername:
Above commit:
image

If its below, then its at the bottom of this list.

bast0006 (approved):
lgtm + works locally

PR #31, opened 2021-07-29
Add issue and pr templates

### LINKED ISSUE #30: Issue and Pull request Templates
Look at the [Github Docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) on how to do this.

onerandomusername:
Closes  #30


.github/ISSUE_TEMPLATE/bug_report_bot.yml
```diff
@@ -0,0 +1,59 @@
...
+labels: ["a: bot", "s: needs triage", "t: bug"]
+assignees: onerandomusername
+body:
+- type: input
+  attributes:
+    label: Is there an existing issue for this?
+    description: Please search to see if an issue already exists for the bug you encountered.
+    placeholder: I have searched the existing issues.
+  validations:
+    required: true
```
bast0006, line 12: This needs a different input type. It should either be a search-for-issues box on the form or a checkbox, and right now it's a text input that gets pasted into the output

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,32 @@
...
+description: Suggest a new feature for the bot and/or website.
+labels: ["s: needs triage", "t: feature"]
+body:
+- type: input
+  attributes:
+    label: Is there an existing issue for this?
+    description: Please search to see if an issue already exists for the bug you encountered.
+    placeholder: I have searched the existing issues.
+  validations:
+    required: true
```
bast0006, line 11: Same deal here as above

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,32 @@
...
+- type: dropdown
+  attributes:
+    label: Would you like to implement this yourself?
+    description: Please note that higher priority features made be added by maintainers.
+    multiple: false
+    options:
+      - Yes, I'd like to implement this
+      - No, someone else can implement it
```
bast0006, line 26: Can we make this a checkbox? Feels awkward

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Discord
+    url: https://discord.gg/YJtPqVntqM
+    about: Please ask and answer questions here. This is also the place to report security vulnerabilities.
```
bast0006, line 5: Can we change `name` to "Questions and help" or something? Questions, concerns, community? We should aim to draw the eyes of people with questions, and `Discord` requires then to think extra (and that's bad, as we're trying to guide low-effort users with this in addition to uncaught cases).

bast0006 (changes requested):
I thought issue forms were more like the markdown templates I'd used before. This is so much better. Let's get these implemented.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+## Relevant Issues
+
+<!--
+It is highly recommended to link to an issue that has been approved by a maintainer, indicated by an approved label.
+Not doing so may result in the pr being closed, without a reason.
```
bast0006, line 5: I don't quite like this phrasing, we ought be kinder here.

> We highly recommend linking to an issue that has been approved by a maintainer, or making one yourself before opening a PR. Otherwise it may not be clear why a chance is necessary or you might do work that we'd already decided not to implement that is not immediately obvious, leading to wasted time and effort.

.github/ISSUE_TEMPLATE/feature_request.yml
onerandomusername: No, if its a checkbox then they don't have to answer it. Its a small thing with the issue forms which annoys me, but you can only make checkboxes themselves required, not the whole question itself.

bast0006 (approved):
lgtm

.github/ISSUE_TEMPLATE/feature_request.yml
bast0006: Ah. Figures. And there doesn't appear to be a radio button option. Is there one? If there is we ought use it, otherwise this is good enough.

PR #43, opened 2021-08-04
feat: extension and plugin loading system

onerandomusername:
## Description

Adds the ability to load extensions, and manage them.

Most of the code of the first commit is taken from [this project](https://github.com/python-discord/bot/tree/a8869b4d60512b173871c886321b261cbc4acca9/bot), at that position in time.

## Added
- Bot modes to determine behaviour. Multiple can be applied at once.
  - `PROD`: the default mode, no dev extensions or dev plugins load
  - `DEVELOP`: the bot developer mode, most useful for people adding features to modmail
    - Enables the extension_manager extension.
  - `PLUGIN_DEV`: the plugin developer mode, useful for enabling plugin-specific features
    - This is not used yet.
- Extension loading system
  - scans for extensions in the `modmail/extensions` folder and loads them if they are of the right format.
- Plugin loading system
  - scans for plugins in the `modmail/plugins` folder and loads them.
- Extension management commands
  - load, reload, unload, list, refresh commands for dealing with extensions
  - Run the `ext` command for more details when the bot is in `DEVELOP` mode.
- Plugin management commands
  - load, reload, unload, list, refresh commands for dealing with plugins
- - Run the `plugins` command for more details.
- Extension metadata
  - used to determine if a cog should load or not depending on the bot mode
- Plugin helper file
  - `modmail/plugin_helpers.py` contains several helpers for making plugins
    - `PluginCog`
    - `ModmailBot`, imported from `modmail.bot`
    - `ModmailLogger`, imported from `modmail.log`
- Meta Cog
  - **NOTE**: The commands in this cog is not stabilized yet and should not be relied upon.
  - Prefix command for getting the set prefix. Most useful by mentioning the bot.
  - Uptime command tells the end-user how long the bot has been online.
  - Ping command to see the bot latency.
- Guide on how to contribute to modmail, see [CONTRIBUTING.md](./CONTRIBUTING.md)
- Start a Changelog
- Add changelog style verify CI check

### Fixed

- Make the bot http_session within an event loop.

## Todo in later issues:
- [ ] manage installing plugins and updating them from Github
- [ ] metadata to determine when in bot start to load extensions (deferred to later pull)
- [ ] metadata to determine dependencies on any other extensions/plugins (deferred to later pull)
- [ ] test suite for all of this new code (deferred to later pull)
- [ ] mock attributes for discord py components (deferred to later pull)


modmail/__init__.py
```diff
@@ -1,17 +1,21 @@
...
 
 logging.TRACE = 5
 logging.NOTICE = 25
 logging.addLevelName(logging.TRACE, "TRACE")
 logging.addLevelName(logging.NOTICE, "NOTICE")
 
-LOG_LEVEL = 20
+LOG_LEVEL = 5
 fmt = "%(asctime)s %(levelname)10s %(name)15s - [%(lineno)5d]: %(message)s"
 datefmt = "%Y/%m/%d %H:%M:%S"
```
Shivansh-007, line 20: Can we capitalize on this?

modmail/__init__.py
```diff
@@ -54,4 +59,5 @@
 # Set asyncio logging back to the default of INFO even if asyncio's debug mode is enabled.
 logging.getLogger("asyncio").setLevel(logging.INFO)
 
-root.debug("Logging initialization complete")
+
+instance: "ModmailBot" = None  # Global ModmailBot instance.
```
Shivansh-007, line 63: Shouldn't the correct annotation be `Optional["ModmailBot"]` since we set it to `None` here?

modmail/__init__.py
```diff
@@ -38,14 +42,15 @@
 
 file_handler.setLevel(logging.TRACE)
 
-coloredlogs.install(
-    level=LOG_LEVEL,
-    fmt=fmt,
-    datefmt=datefmt,
-)
+# configure trace color
+_LEVEL_STYLES = dict(coloredlogs.DEFAULT_LEVEL_STYLES)
```
Shivansh-007, line 46: Can we make this configurable as `env` variable, it really helps when someone is using a weird colour scheme.

modmail/__init__.py
```diff
@@ -38,14 +42,15 @@
 
 file_handler.setLevel(logging.TRACE)
 
-coloredlogs.install(
-    level=LOG_LEVEL,
-    fmt=fmt,
-    datefmt=datefmt,
-)
+# configure trace color
```
Shivansh-007, line 45: Couldn't put this anywhere else but rather than the relational path of `log_file` wouldn't be better to use `Path("logs", "bot.log")`

modmail/config.py
```diff
@@ -152,6 +152,9 @@ class DevConfig(BaseSettings):
     """
 
     log_level: conint(ge=0, le=50) = getattr(logging, "NOTICE", 25)
+    production: bool = True
+    devel: bool = False
```
Shivansh-007, line 156: devil!
```suggestion
    development: bool = False
```

modmail/exts/meta.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Meta(commands.Cog):
+    """Meta commands to get info about the bot itself."""
+
+    def __init__(self, bot: ModmailBot):
+        self.bot = bot
+
+    @commands.command()
+    async def ping(self, ctx: Context) -> None:
+        """Check response time."""
```
Shivansh-007, line 22: Better, thanks alec!
```suggestion
        """Ping the bot to see its latency and state."""
```

modmail/exts/meta.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Meta(commands.Cog):
+    """Meta commands to get info about the bot itself."""
+
+    def __init__(self, bot: ModmailBot):
+        self.bot = bot
+
+    @commands.command()
+    async def ping(self, ctx: Context) -> None:
```
Shivansh-007, line 21: `pong` alias, please.

modmail/exts/meta.py
```diff
@@ -0,0 +1,34 @@
+import logging
+
+from discord.ext import commands
+from discord.ext.commands import Context
+
+from modmail.bot import ModmailBot
+from modmail.log import ModmailLogger
+from modmail.utils.cogs import CogMetadata
```
Shivansh-007, line 8: I don't think this even gets used anywhere, same for line 11.

modmail/utils/cogs.py
```diff
@@ -0,0 +1,36 @@
...
+
+class BotModes(IntEnum):
+    """
+    Valid modes for the bot.
+
+    These values affect logging levels, which logs are loaded, and so forth.
+    """
+
+    production = int("1", 2)
+    devel = int("10", 2)
```
Shivansh-007, line 13: devil!
```suggestion
    development = int("10", 2)
```

modmail/utils/cogs.py
```diff
@@ -0,0 +1,36 @@
...
+BOT_MODES = BotModes
+
+
+@dataclass()
+class CogMetadata:
+    """Cog metadata class to determine if cog should load at runtime depending on bot configuration."""
+
+    # load if bot is in development mode
+    # development mode is when the bot has its metacogs loaded, like the eval and extension cogs
+    devel: bool = False
```
Shivansh-007, line 26: devil!
```suggestion
    development: bool = False
```

Shivansh-007 (changes requested):

modmail/exts/meta.py
onerandomusername: This is used in the cog metadata system, where its not used in the file but is imported and checked in modmail.utils.extensions to see if the file should load. 

However, this was removed since I modified the system to load by default if this does not exist, and if it is a develop or plugin extension then it should have this imported and the specific variables set.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+
+    async def convert(self, ctx: Context, argument: str) -> str:
+        """Fully qualify the name of an extension and ensure it exists."""
+        # Special values to reload all extensions
+        if argument == "*" or argument == "**":
+            return argument
+
+        argument = argument.lower()
+        extensions = []
+        for ext, _nul in EXTENSIONS:
```
Shivansh-007, line 54: This seems to be getting repeated at a few places, can we take this to the top of the file  as `EXTENSION_NAMES` or something like that? Since this is just a for loop, it can be ignored but I had like this change.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+    async def list_command(self, ctx: Context) -> None:
+        """
+        Get a list of all extensions, including their loaded status.
+
+        Grey indicates that the extension is unloaded.
+        Green indicates that the extension is currently loaded.
+        """
+        embed = Embed(colour=Colour.blurple())
+        embed.set_author(
+            name="Extensions List",
```
Shivansh-007, line 167: Could you add `icon_url` and maybe a link to the `ext/` folder to this.

onerandomusername:
Still pending writing tests, but the rest of the pull looks ready.

onerandomusername:
Still pending writing tests, but the rest of the pull looks ready.

modmail/__init__.py
Shivansh-007: This hasn't been mentioned yet.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+        """Return a mapping of extension names and statuses to their categories."""
+        categories = {}
+        extensions = []
+        for ext, _nul in EXTENSIONS:
+            extensions.append(ext)
+        for ext in extensions:
+            if ext in self.bot.extensions:
+                status = ":thumbsup:"
+            else:
+                status = ":thumbsdown:"
```
Shivansh-007, line 193: I feel like these emojis are big and when placed one below the other then merge together, are there are small other emojis, maybe a `red` and `green` dot or something would also work pretty well.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+        else:
+            msg = f":ok_hand: Extension successfully {verb}ed: `{ext}`."
+            log.debug(msg[10:])
+
+        return msg, error_msg
+
+    # This cannot be static (must have a __func__ attribute).
+    async def cog_check(self, ctx: Context) -> bool:
+        """Only allow bot owners to invoke the commands in this cog."""
+        return await self.bot.is_owner(ctx.author)
```
Shivansh-007, line 266: I think even admins should be allowed to do this.

modmail/exts/meta.py
```diff
@@ -0,0 +1,37 @@
...
+    @commands.command(name="uptime")
+    async def uptime(self, ctx: commands.Context) -> None:
+        """Get the current uptime of the bot."""
+        timestamp = self.bot.start_time.format("X").split(".")[0]
+        await ctx.send(f"Start time: <t:{timestamp}:R>")
+
+    @commands.command(name="prefix")
+    async def prefix(self, ctx: commands.Context) -> None:
+        """Return the configured prefix."""
+        await ctx.send(f"My current prefix is `{self.bot.config.bot.prefix}`")
```
Shivansh-007, line 32: Uhhhhhhh, can't we just put this in the bot's playing game status `Commands: {prefix}help`?

Shivansh-007 (changes requested):
Same comments for plugin manager as for the extensions cog.

I am still not convinced with the approach as plugin and config have really really same code, I had like to merge those two with the help of util functions maybe like `modmail/ext/extensions/plugin.py` and `extensions.py` with both **inheriting** the main class which is situated in `__init__.py`.

modmail/exts/meta.py
onerandomusername: No, since this is a modmail bot, the playing is typically a dm for help status. Also, the majority of server members don't need to know said prefix.

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+EXT_METADATA = ExtMetadata(production=True, develop=True, plugin_dev=True)
+
+
+class Action(Enum):
+    """Represents an action to perform on an extension."""
+
+    # Need to be partial otherwise they are considered to be function definitions.
+    LOAD = functools.partial(ModmailBot.load_extension)
+    UNLOAD = functools.partial(ModmailBot.unload_extension)
+    RELOAD = functools.partial(ModmailBot.reload_extension)
```
bast0006: Don't define this twice. AS above, this should be a dict. Define it on the ModmailCog class, and then have plugins subclass that.

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class Plugin(commands.Converter):
```
bast0006, line 37: PluginConverter

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        If '\*' or '\*\*' is given as the name, all unloaded extensions will be loaded.
+        """  # noqa: W605
+        if not plugs:
+            await ctx.send_help(ctx.command)
+            return
+
+        if "*" in plugs or "**" in plugs:
+            plugs = set(PLUGINS) - set(self.bot.extensions.keys())
+
+        msg = self.batch_manage(Action.LOAD, *plugs)
```
bast0006, line 103: sorted()

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        Unload currently loaded extensions given their fully qualified or unqualified names.
+
+        If '\*' or '\*\*' is given as the name, all loaded extensions will be unloaded.
+        """  # noqa: W605
+        if not plugs:
+            await ctx.send_help(ctx.command)
+            return
+
+        if "*" in plugs or "**" in plugs:
+            plugs = set(self.bot.extensions.keys())
```
bast0006, line 118: no need for set() here, and it mangles the natural load order

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        if not plugs:
+            await ctx.send_help(ctx.command)
+            return
+
+        if "**" in plugs:
+            plugs = []
+            for plug, _nul in PLUGINS:
+                plugs.append(plug)
+        elif "*" in plugs:
+            plugs = set(self.bot.extensions.keys()) | set(plugs)
```
bast0006, line 143: Same as above

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        If '\*' is given as the name, all currently loaded extensions will be reloaded.
+        If '\*\*' is given as the name, all extensions, including unloaded ones, will be reloaded.
+        """  # noqa: W605
+        if not plugs:
+            await ctx.send_help(ctx.command)
+            return
+
+        if "**" in plugs:
+            plugs = []
+            for plug, _nul in PLUGINS:
```
bast0006, line 140: PLUGINS ought be a dict here too

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        embed.set_author(
+            name="Plugins List",
+        )
+
+        lines = []
+        categories = self.group_extension_statuses()
+        for category, plugs in sorted(categories.items()):
+            # Treat each category as a single line by concatenating everything.
+            # This ensures the paginator will not cut off a page in the middle of a category.
+            category = category.replace("_", " ").title()
```
bast0006, line 168: Perhaps it's time for a utility function? `Format Category`?

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+            plugs = "\n".join(sorted(plugs))
+            lines.append(f"**{category}**\n{plugs}\n")
+
+        log.debug(f"{ctx.author} requested a list of all cogs. Returning a list.")
+        # await Paginator.paginate(lines, ctx, embed, scale_to_size=700, empty=False)
+
+        # since we currently don't have a paginator.
+        output = ""
+        for line in lines:
+            output += line
```
bast0006, line 178: "".join()

bast0006: If you want, make a paginator dummy class that just does this

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+            if len(path) > BASE_PATH_LEN + 1:
+                category = " - ".join(path[BASE_PATH_LEN:-1])
+            else:
+                category = "uncategorised"
+
+            categories.setdefault(category, []).append(f"{status}  {path[-1]}")
+
+        return categories
+
+    def batch_manage(self, action: Action, *plugs: str) -> str:
```
bast0006, line 203: I don't like this duped code. We can probably treat plugins as subclasses of extensions

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            raise commands.BadArgument(f":x: Could not find the plugin `{argument}`.")
+
+
+class PluginManager(commands.Cog, name="Plugin Manager"):
+    """Plugin management commands."""
+
+    def __init__(self, bot: ModmailBot):
+        self.bot = bot
+
```
bast0006, line 83: Subclass ExtensionManager and just reuse the code, sans formatting stuff, with super()?

modmail/utils/cogs.py
```diff
@@ -0,0 +1,40 @@
...
+class BotModes(IntEnum):
+    """
+    Valid modes for the bot.
+
+    These values affect logging levels, which logs are loaded, and so forth.
+    """
+
+    production = int("1", 2)
+    develop = int("10", 2)
+    plugin_dev = int("100", 2)
```
bast0006, line 14: 1>>0
1>>1
1>>2 
is neater

modmail/utils/cogs.py
```diff
@@ -0,0 +1,40 @@
...
+    develop: bool = False
+    # plugin development mode
+    # used for loading bot plugins that help with plugin debugging
+    plugin_dev: bool = False
+
+
+def calc_mode(metadata: ExtMetadata) -> int:
+    """Calculate the combination of different variables and return the binary combination."""
+    mode = int(getattr(metadata, "production", False))
+    mode = mode + int(getattr(metadata, "develop", False) << 1) or 0
```
bast0006, line 38: +=

bast0006: Also no need for getattr here unless we want to meta-code this. ( for name in ...) which we shouldn't. We should be directly translating using the enum members values and names. We can define `def _generate_next_value_(name, start, count, last_values):` to autogenerate these and just provide names, and `__members__` to determine what possible values there are... in order, and then we don't have to hardcode any of this

modmail/utils/extensions.py
```diff
@@ -0,0 +1,62 @@
...
+            log.trace(f"Load cog {module.name!r}?: {load_cog}")
+            yield module.name, load_cog
+            continue
+
+        log.notice(f"Cog {module.name!r} is missing an EXT_METADATA variable. Assuming its a normal cog.")
+
+        yield (module.name, True)
+
+
+EXTENSIONS = frozenset(walk_extensions())
```
bast0006, line 62: dict()

modmail/utils/plugin_manager.py
```diff
@@ -0,0 +1,67 @@
...
+            continue
+
+        log.notice(
+            f"Plugin {imported.__name__!r} is missing a EXT_METADATA variable. Assuming its a normal plugin."
+        )
+
+        yield (imported.__name__, True)
+
+
+PLUGINS = frozenset(walk_plugins())
```
bast0006, line 67: This is duped code. There's gotta be a better way, maybe call extension loading form the extensions tool through here?

bast0006 (changes requested):
big stuff: PLUGINS and EXTENSIONS should be dicts, or just lists with their enabled status as a class attribute.

Loading/determining if something is a plugin/extension should not be module wide but class wide, determined by subclassing a Plugin/ModmailExtension/ModmailBot class. 

Make sure to sorted() stuff before performing mass actions for consistency

Extension metadata should be on extension class

modmail/utils/cogs.py
bast0006: Actually:
```py
class BitwiseAutoEnum(IntEnum):
  def _generate_next_value_(name, start, count, last_values):
    return 1>>count
```

modmail/utils/cogs.py
bast0006: See comment above this set with BitwiseAutoEnum, you can list() an enum to get it's members so use that and .name/.value to do this magically

Procfile
```diff
@@ -1 +1 @@
-worker: python -m modmail
+web: python -m modmail
```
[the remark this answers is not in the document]
onerandomusername: Yeah, unfortunately.

modmail/utils/cogs.py
onerandomusername: How does that work?

modmail/utils/plugin_manager.py
onerandomusername: If we call anything in that file we run the risk of recursive imports. What if we copied some of these methods to another utils file?

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+
+log: ModmailLogger = logging.getLogger(__name__)
+
+
+UNLOAD_BLACKLIST = {
+    __name__,
+}
+BASE_PATH_LEN = len(exts.__name__.split("."))
+
+EXT_METADATA = ExtMetadata(develop=True)
```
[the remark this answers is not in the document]
onerandomusername: Not all extensions are necessarily a cog.

modmail/utils/cogs.py
pyxiis: @onerandomusername aren't these just logical shifts?

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+from modmail.utils.cogs import ExtMetadata
+from modmail.utils.extensions import EXTENSIONS, unqualify
+
+log: ModmailLogger = logging.getLogger(__name__)
+
+
+UNLOAD_BLACKLIST = {
+    __name__,
+}
+BASE_PATH_LEN = len(exts.__name__.split("."))
```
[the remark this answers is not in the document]
Shivansh-007: Um, I think this is more clear to what exactly is the variable being referred to.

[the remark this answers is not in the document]
onerandomusername: @Shivansh-007 
```suggestion
BASE_PATH_LEN = exts.__name__.count(".") +1
```

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,278 @@
...
+            return argument
+
+        argument = argument.lower()
+        extensions = []
+        for ext, _nul in EXTENSIONS:
+            extensions.append(ext)
+
+        if argument in extensions:
+            return argument
+        elif (qualified_arg := f"{exts.__name__}.{argument}") in EXTENSIONS:
```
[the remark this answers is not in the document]
Shivansh-007: We could, but I don't see any reason to do so, since both are doing the same main task.

modmail/exts/utils/plugin_manager.py
onerandomusername: Importing ExtensionManager from PluginManager will cause a recursive import when utils.extensions looks at exts.utils.plugin_manager

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,282 @@
...
+EXT_METADATA = ExtMetadata(develop=True)
+
+
+class Action(Enum):
+    """Represents an action to perform on an extension."""
+
+    # Need to be partial otherwise they are considered to be function definitions.
+    LOAD = functools.partial(ModmailBot.load_extension)
+    UNLOAD = functools.partial(ModmailBot.unload_extension)
+    RELOAD = functools.partial(ModmailBot.reload_extension)
```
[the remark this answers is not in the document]
Shivansh-007: If we wanted to access them dynamically I would have used a dictionary here, but that's not the case. And I think our use fits the [PEP 435](https://www.python.org/dev/peps/pep-0435/) perfectly,

> The properties of an enumeration are useful for defining an immutable, related set of constant values that may or may not have a semantic meaning.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,282 @@
...
+
+    async def convert(self, ctx: Context, argument: str) -> str:
+        """Fully qualify the name of an extension and ensure it exists."""
+        # Special values to reload all extensions
+        if argument == "*" or argument == "**":
+            return argument
+
+        argument = argument.lower()
+        extensions = []
+        for ext, _nul in EXTENSIONS:
```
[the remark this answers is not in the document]
Shivansh-007: Agreed.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,282 @@
...
+        verb = action.name.lower()
+        failures = {}
+
+        for extension in extensions:
+            _, error = self.manage(action, extension)
+            if error:
+                failures[extension] = error
+
+        emoji = ":x:" if failures else ":ok_hand:"
+        msg = f"{emoji} {len(extensions) - len(failures)} / {len(extensions)} extensions {verb}ed."
```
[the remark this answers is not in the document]
Shivansh-007: They aren't being really divided? it's just shown as `2 / 3`?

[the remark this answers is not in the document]
onerandomusername: Yeah, this is an f-string, although for readability, a `.format()` would probably be better.

Shivansh-007:
@onerandomusername @bast0006 What do you think of my review comment, of merging them into a base class and the two inheriting them?

bast0006:
> @onerandomusername @bast0006 What do you think of my review comment, of merging them into a base class and the two inheriting them?

Yep. That's what I was thinking as well and mentioned in the wall of review somewhere. Honestly I think we don't need a base class, Plugin should just subclass Extension.

modmail/__init__.py
```diff
@@ -31,21 +37,22 @@
...
+LEVEL_STYLES = dict(coloredlogs.DEFAULT_LEVEL_STYLES)
```
Shivansh-007, line 48: Could we also make these configurable, environment variables?

modmail/__main__.py
```diff
@@ -8,15 +8,17 @@
 except ImportError:
     pass
 
-from .bot import ModmailBot
+from modmail.bot import ModmailBot
 
 log = logging.getLogger(__name__)
```
Shivansh-007, line 13: Missing the `ModmailLogger` typehint.

modmail/bot.py
```diff
@@ -19,51 +16,85 @@ class ModmailBot(commands.Bot):
     """
 
     main_task: asyncio.Task
+    logger = logging.getLogger(__name__)
 
     def __init__(self, **kwargs):
         self.config = CONFIG
         self.internal = INTERNAL
         self.http_session: ClientSession = None
```
Shivansh-007, line 24:
```suggestion
        self.http_session: t.Optional[ClientSession] = None
```
Do add `import typing as t` at the start of the file.

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class ExtensionConverter(commands.Converter):
+    """
+    Fully qualify the name of an extension and ensure it exists.
+
+    The * and ** values bypass this when used with the reload command.
+    """
+
+    source_list = EXTENSIONS
```
Shivansh-007, line 43:
```suggestion
    source_list = EXTENSIONS.copy()
```

The assignment just copies the reference to the list, not the actual list, so both `source_list` and `EXTENSIONS` refer to the same list after the assignment.

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,276 @@
...
+    type = "extension"
+
+    async def convert(self, ctx: Context, argument: str) -> str:
```
Shivansh-007, line 46:
```suggestion
    async def convert(self, _: Context, argument: str) -> str:
```

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        msg = self.batch_manage(Action.LOAD, *extensions)
+        await ctx.send(msg)
+
+    @extensions_group.command(name="unload", aliases=("ul",))
+    async def unload_extensions(self, ctx: Context, *extensions: ExtensionConverter) -> None:
+        """
+        Unload currently loaded extensions given their fully qualified or unqualified names.
+
+        If '\*' or '\*\*' is given as the name, all loaded extensions will be unloaded.
+        """  # noqa: W605
```
Shivansh-007, line 120: Not needed.

```suggestion
        """
        Unload currently loaded extensions given their fully qualified or unqualified names.

        If '*' or '**' is given as the name, all loaded extensions will be unloaded.
        """
```

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,276 @@
...
+        """Load, unload, reload, and list loaded extensions."""
+        await ctx.send_help(ctx.command)
+
+    @extensions_group.command(name="load", aliases=("l",))
+    async def load_extensions(self, ctx: Context, *extensions: ExtensionConverter) -> None:
+        """
+        Load extensions given their fully qualified or unqualified names.
+
+        If '\*' or '\*\*' is given as the name, all unloaded extensions will be loaded.
+        """  # noqa: W605
```
Shivansh-007, line 103: Not needed.

```suggestion
        """
        Load extensions are given their fully qualified or unqualified names.

        If '*' or '**' is given as the name, all unloaded extensions will be loaded.
        """
```

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,276 @@
...
+            if ext in self.bot.extensions:
+                status = ":green_circle:"
+            else:
+                status = ":red_circle:"
+
+            root, name = ext.rsplit(".", 1)
+
+            category = " - ".join(root.split("."))
+
+            categories[category].append(f"{status}  {name}")
```
Shivansh-007, line 200:
```suggestion
            root, name = ext.rsplit(".", 1)
            category = " - ".join(root.split("."))
            categories[category].append(f"{status}  {name}")
```

modmail/extensions/plugin_manager.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @plugins_group.command(name="unload", aliases=("ul",))
+    async def unload_plugins(self, ctx: Context, *plugins: PluginConverter) -> None:
+        """
+        Unload currently loaded plugins given their fully qualified or unqualified names.
+
+        If '\*' or '\*\*' is given as the name, all loaded plugins will be unloaded.
+        """  # noqa: W605
+        await self.unload_extensions.callback(self, ctx, *plugins)
+
```
Shivansh-007, line 54:
```suggestion

    @plugins_group.command(name="load", aliases=("l",))
    async def load_plugin(self, ctx: Context, *plugins: PluginConverter) -> None:
        """
        Load plugins given their fully qualified or unqualified names.

        If '*' or '**' is given as the name, all unloaded plugins will be loaded.
        """  # noqa: W605
        await self.load_extensions.callback(self, ctx, *plugins)

    @plugins_group.command(name="unload", aliases=("ul",))
    async def unload_plugins(self, ctx: Context, *plugins: PluginConverter) -> None:
        """
        Unload currently loaded plugins given their fully qualified or unqualified names.

        If '*' or '**' is given as the name, all loaded plugins will be unloaded.
        """  # noqa: W605
        await self.unload_extensions.callback(self, ctx, *plugins)

```

Shivansh-007 (changes requested):
Nitpicks

modmail/extensions/extension_manager.py
onerandomusername: It's needed due to markdown,  as the docstring is used as the help info

modmail/extensions/extension_manager.py
onerandomusername: Due to discord markdown, this is needed.

modmail/extensions/extension_manager.py
onerandomusername: Due to discord markdown, this is needed.

modmail/extensions/plugin_manager.py
onerandomusername: See above.

modmail/exts/utils/extensions.py
onerandomusername: Part of the reason to leave this as bot owner only is because in some cases, owners don't uncheck the box "make my bot public" because *for some reason* discord decides to leave that on for all new bots..

modmail/exts/meta.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def setup(bot: ModmailBot) -> None:
+    """Load the Meta cog."""
+    bot.add_cog(Meta(bot))
```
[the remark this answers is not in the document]
onerandomusername: Resolved in discord.

tl;dr discord.py's load extension method searches for a setup method, since extensions could very well just be a single command, cogs are just easy ways to make groups of commands in a structure.

modmail/bot.py
```diff
@@ -64,6 +69,15 @@ def add_cog(self, cog: commands.Cog) -> None:
...
 
+    def remove_cog(self, cog: commands.Cog) -> None:
+        """
+        Delegate to super to unregister `cog`.
+
+        This only serves to make the debug log, so that extensions don't have to.
+        """
+        super().remove_cog(cog)
+        log.trace(f"Cog unloaded: {cog}")
+
```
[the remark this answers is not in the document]
onerandomusername: Base class now exists.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,282 @@
...
+
+        return msg
+
+    def manage(self, action: Action, ext: str) -> t.Tuple[str, t.Optional[str]]:
+        """Apply an action to an extension and return the status message and any error message."""
+        verb = action.name.lower()
+        error_msg = None
+
+        try:
+            action.value(self.bot, ext)
```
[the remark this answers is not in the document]
onerandomusername: Resolved, keeping it an enum.

modmail/__main__.py
```diff
@@ -17,7 +17,9 @@ def main() -> None:
     """Run the bot."""
     bot = ModmailBot()
     log.notice("running bot")
-    bot.run(bot.config.bot.token)
+    bot.instance = ModmailBot()
+    bot.instance.load_extensions()
+    bot.instance.run(bot.config.bot.token)
```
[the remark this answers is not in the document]
onerandomusername: What do you mean?

modmail/exts/utils/plugin_manager.py
```diff
@@ -0,0 +1,276 @@
...
+from modmail.log import ModmailLogger
+from modmail.utils.cogs import ExtMetadata
+from modmail.utils.plugin_manager import PLUGINS, unqualify
+
+log: ModmailLogger = logging.getLogger(__name__)
+
+BASE_PATH = Path(plugins.__file__).parent
+BASE_PATH_LEN = len(plugins.__name__.split("."))
+
+EXT_METADATA = ExtMetadata(production=True, develop=True, plugin_dev=True)
```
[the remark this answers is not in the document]
onerandomusername: Due to the fact that not every extension is technically a cog, This needs to be a top level variable.

modmail/extensions/extension_manager.py
onerandomusername: I think its intended behaviour.

modmail/exts/utils/extensions.py
```diff
@@ -0,0 +1,282 @@
...
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            if action is Action.RELOAD:
+                # When reloading, just load the extension if it was not loaded.
+                log.debug("Treating {ext!r} as if it was not loaded.")
+                return self.manage(Action.LOAD, ext)
```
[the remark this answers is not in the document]
onerandomusername: I don't think this hides anything.
image

modmail/__init__.py
bast0006: I suppose that could be possible but it is an irritatingly detailed dict structure and I didn't think it was worth it.

modmail/extensions/extension_manager.py
bast0006: We don't want to copy this. We should keep this list as a reference, not a copy, to ensure that extensions added or removed via other methods remain in sync with the source list

modmail/__init__.py
onerandomusername: If we don't set this variable, coloredlogs actually will use a environment variable as override. So we could check if that env var is set, and if it is, do *not* set our level styles.

onerandomusername: Set up, install the optionals `environs` to use this feature

modmail/extensions/extension_manager.py
Shivansh-007: Not mentioned...

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,278 @@
...
+            raise commands.BadArgument(
+                f":x: `{argument}` is an ambiguous {self.type} name. "
+                f"Please use one of the following fully-qualified names.```\n{names}```"
+            )
+
+        return matches[0]
+
+
+class ExtensionManager(ModmailCog, name="Extension Manager"):
+    """Extension management base class."""
```
Shivansh-007, line 80: It is not really a base class and is a functional cog itself but a base class for the plugins. So I am not sure if this is a perfect docstring. Maybe you can clear it up :D

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,278 @@
...
+    @extensions_group.command(name="load", aliases=("l",))
+    async def load_extensions(self, ctx: Context, *extensions: ExtensionConverter) -> None:
+        """
+        Load extensions given their fully qualified or unqualified names.
+
+        If '\*' or '\*\*' is given as the name, all unloaded extensions will be loaded.
+        """  # noqa: W605
+        if not extensions:
+            await ctx.send_help(ctx.command)
+            return
```
Shivansh-007, line 106: This is against our guru the holy PEP8, if this isn't solved bad deeds will happen with modmail 🆘 

```suggestion
            return None
```

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as `return None`, and an explicit return statement should be present at the end of the function (if reachable)

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,278 @@
...
+        for category, extensions in sorted(categories.items()):
+            # Treat each category as a single line by concatenating everything.
+            # This ensures the paginator will not cut off a page in the middle of a category.
+            category = category.replace("_", " ").title()
+            extensions = "\n".join(sorted(extensions))
+            lines.append(f"**{category}**\n{extensions}\n")
+
+        log.debug(f"{ctx.author} requested a list of all {self.type}s. " "Returning a paginated list.")
+
+        # since we currently don't have a paginator.
```
Shivansh-007, line 185: Add a TODO mark here, so we know to change this later.

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,278 @@
...
+        await ctx.send("".join(lines) or f"( There are no {self.type}s installed. )")
```
Shivansh-007, line 186: We should check if no types are installed before we start looping through them to form the message. It is more clear to what it is exactly doing and maybe even saves negligible performance.

modmail/extensions/extension_manager.py
Shivansh-007: Huh? I just removed the empty lines in between to group them into a single section as they are one.

modmail/extensions/plugin_manager.py
```diff
@@ -0,0 +1,104 @@
...
+    # TODO: Implement install/enable/disable/etc
+
+    # This cannot be static (must have a __func__ attribute).
+    async def cog_check(self, ctx: Context) -> bool:
+        """Only allow server admins and bot owners to invoke the commands in this cog."""
+        return ctx.author.guild_permissions.administrator or await self.bot.is_owner(ctx.author)
+
+
+# Delete the commands from ExtensionManager that PluginManager has inherited
```
Shivansh-007, line 95: Add a `HACK:` mark, in case we figure out a better way to this inheritance in future.

pyproject.toml
```diff
@@ -18,17 +18,26 @@ packages = [{ include = "modmail" }]
 [tool.poetry.dependencies]
 python = "^3.8"
 aiohttp = { extras = ["speedups"], version = "^3.7.4" }
+arrow = "^1.1.1"
 colorama = "^0.4.3"
 coloredlogs = "^15.0"
 "discord.py" = {git = "https://github.com/Rapptz/discord.py.git", rev = "master"}
+environs = {version = "~=9.3.3", optional = true}
 pydantic = "^1.8.2"
 python-dotenv = "~=0.17.1"
```
Shivansh-007, line 27: Why are we using two env loaders here? Choose one, I had say to go with `environs` until we get configurave in a stable state.

Shivansh-007 (changes requested):

modmail/extensions/extension_manager.py
onerandomusername: If lines is empty, it'll be a blank return and instantly go to the next. For the amount of times there *will* be something installed, this is faster overall.

modmail/extensions/extension_manager.py
onerandomusername: > Either all return statements in a function should return an expression, or none of them should.

None of the other return statements in this function return a value either.

modmail/extensions/extension_manager.py
bast0006: If we are going to check for installed before we create the message (which is reasonable) then it should be a guard statement at the top of the function.

pyproject.toml
onerandomusername: I do not want to touch the config file until we have a different thing for it, but if you insist...

modmail/config.py
```diff
@@ -145,13 +145,26 @@ class ChannelConfig(BaseSettings):
...
+class BotMode(BaseSettings):
+    """
+    Bot mode.
+
+    Used to determine when the bot will run.
+    """
+
+    production: bool = True
+    plugin_dev: bool = False
+    develop: bool = False
```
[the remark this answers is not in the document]
onerandomusername: They are not optional due to the fact that the values they are assigned here are their default values. Also this file is going to be gutted anyways and replaces with a different configuration system.

pyproject.toml
bast0006: I think we ought leave this for cleanup until we make a configurave integration pr

Shivansh-007: Perfect

modmail/extensions/extension_manager.py
Shivansh-007: > If lines is empty, it'll be a blank return and instantly go to the next. For the amount of times there will be something installed, this is faster overall.

Hence, I said it as ineligible performance and focused more on clarity of the functionality, but that's my opinion and the reason bast said.

modmail/config.py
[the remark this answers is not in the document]
Shivansh-007: I don't see any other reason to assign default values to them if there are not optional values, as they are gonna be overwritten anyway.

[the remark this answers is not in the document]
onerandomusername: They're required for bot function, optional for the user to put in.

modmail/extensions/extension_manager.py
Shivansh-007: I had take it as a complete decision though to maintain consistency throughout the repository. Either we keep it or not, doesn't really matter for me.

modmail/exts/utils/extensions.py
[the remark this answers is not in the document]
bast0006: What happens if an extension writes to a file on unload but crashes before that happens, and on load it attempts to read from the file? You'll get two errors on a reload, a failed unload, and then a failed load. We should instead fail early so this can be handled right rather than potentially damage extension state by performing two loads without an unload in between.

bast0006 (changes requested):

Shivansh-007 (approved):

modmail/exts/utils/extensions.py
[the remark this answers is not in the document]
onerandomusername: Good point, I've removed the code that enabled this to happen.

The code now does not intentionally catch these errors, and sends a message if cogs are already loaded or not.

onerandomusername (commented):
One change to make:

Changelog needs to be updated to 0.1.0. This will happen last, after a review.

ShakyaMajumdar:
The `.env.template` file should have descriptions for the variables not mentioned in CONTRIBUTING.md.

onerandomusername:
> 
> 
> The `.env.template` file should have descriptions for the variables not mentioned in CONTRIBUTING.md.

Addressed by removing vars which the bot does not currently use.

CONTRIBUTING.md
```diff
@@ -0,0 +1,239 @@
...
+  - **Explain why this enhancement would be useful** to Modmail, and would benefit the community members.
+- Please try and be clear about why existing features and alternatives would not work for you.
+
+Once it's filed:
+
+- The Maintainers will [label the issue](#label-issues).
+- The Maintainers will evaluate the feature request, possibly asking you more questions to understand its purpose and any relevant requirements. If the issue is closed, the team will convey their reasoning and suggest an alternative path forward.
+- If the feature request is accepted, it will be marked for implementation with `status: approved`, which can then be done either by a core team member or by anyone in the community who wants to contribute code.
+
+> **Note**: The team is unlikely to be able to accept every single feature request that is filed. Please understand if they need to say no.
```
bast0006, line 70: Mention being able to write a plugin again here. Best to have no chances of missing that option rather than field extra issues.

CONTRIBUTING.md
```diff
@@ -0,0 +1,239 @@
...
+So you want to contribute some code! That's great! This project uses GitHub Pull Requests to manage contributions, so [read up on how to fork a GitHub project and file a PR](https://guides.github.com/activities/forking) if you've never done it before.
+
+### Test Server and Bot Account
+
+You will need your own test server and bot account on Discord to test your changes to the bot.
+
+1. Create a test server.
+1. Create a bot account and invite it to the server you just created.
+
+Note down the IDs for your server, as well as any channels and roles created.
```
bast0006, line 83: Should put in a TODO of which roles we might need.. if any? Do we need any roles? Should we just mention "custom roles" because we interact with those?

CONTRIBUTING.md
```diff
@@ -0,0 +1,239 @@
...
+To contribute code:
+
+- [Set up the project](#project-setup).
+- Make any necessary changes to the source code.
+- Write clear, concise commit message(s).
+  - A more in-depth guide to writing great commit messages can be found in Chris Beam's [*How to Write a Git Commit Message*](https://chris.beams.io/posts/git-commit/).
+- Run `flake8`, `black` and `pre-commit` against your code **before** you push. Your commit will be rejected by the build server if it fails to lint. You can run the lint by executing `poetry run task lint` in your command line.
+- Go to [discord-modmail/modmail/pulls][modmail-pulls] and open a new pull request with your changes.
+- If PRing from your own fork, **ensure that "Allow edits from maintainers" is checked**. This permits maintainers to commit changes directly to your fork, speeding up the review process.
+- If your PR is connected to an open issue, add a line in your PR's description that says `Closes #123`, where `#123` is the number of the issue you're fixing.
```
bast0006, line 167: ... (This will make github link your issue and make it easier for us and other contributers to find and understand your PR )

modmail/bot.py
```diff
@@ -19,51 +20,107 @@ class ModmailBot(commands.Bot):
...
+        for plugin, should_load in PLUGINS.items():
+            if should_load:
+                self.logger.debug(f"Loading plugin {plugin}")
+                try:
+                    # since we're loading user generated content,
+                    # any errors here will take down the entire bot
+                    self.load_extension(plugin)
+                except Exception:
+                    self.logger.error("Failed to load plugin {0}".format(plugin), exc_info=True)
```
bast0006, line 96: In the future, we could consider splitting filling the list and loading plugins separately if we want. I've also got an itch that this'll cause nasty issues if stuff gets loaded before this into plugins (should never happen, so not worth worrying about)

Feel free to insta-resolve this, just commentary.

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,301 @@
...
+
+        If an extension fails to be reloaded, it will be rolled-back to the prior working state.
+
+        If '\*' or '\*\*' is given as the name, all currently loaded extensions will be reloaded.
+        """  # noqa: W605
+        if not extensions:
+            await ctx.send_help(ctx.command)
+            return
+
+        if "*" in extensions or "**" in extensions:
```
bast0006, line 159: Ditto. Don't forget the docstring

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,301 @@
...
+
+        Typical use case is in the event that the existing extensions have changed while the bot is running.
+        """
+        log.debug(f"Refreshing list of {self.type}s.")
+
+        # make sure the new walk contains all currently loaded extensions, so they can be unloaded
+        loaded_extensions = dict()
+        for ext in self.all_extensions.items():
+            if ext in self.bot.extensions:
+                loaded_extensions.update(ext)
```
bast0006, line 204: This always fails. `.items()` returns a typle, `bot.extensions` is a dictionary of `name: extension`. A tuple will never be in a list like that, so this will never treat any extension as loaded. Instead, we should just use nothing, I believe, as that ought be more appropriate. 

```py
for ext_name, should_load in self.all_extensions.items():
    if ext_name in self.bot.extensions:
        loaded_extensions[ext_name] = should_load
```

Also.. is it possible for us to write this as a generator expression passing it into loaded_extensions?

modmail/extensions/extension_manager.py
```diff
@@ -0,0 +1,301 @@
...
+        # now that we know what the list was, we can clear it
+        self.all_extensions.clear()
+        # put the loaded extensions back in
+        self.all_extensions.update(loaded_extensions)
+        # now we can re-walk the extensions
+        self.all_extensions.update(self.refresh_method())
+        await ctx.send(f":ok_hand: Refreshed list of {self.type}s.")
+
```
bast0006, line 212: This needs a comment explaining why we do this and what it's for. This is "reload extensions from disk, but do not unload any currently active", but it's not obvious on first glance.

modmail/utils/extensions.py
```diff
@@ -0,0 +1,79 @@
...
+
+def determine_bot_mode() -> int:
+    """
+    Figure out the bot mode from the configuration system.
+
+    The configuration system uses true/false values, so we need to turn them into an integer for bitwise.
+    """
+    bot_mode = 0
+    for mode in BotModes:
+        if getattr(CONFIG.dev.mode, str(mode).split(".")[-1].lower(), True):
```
bast0006, line 32: This needs a comment. I have no idea what this does. I apologize for being a hypocrite, but this deserves like.. a 3 line explaining comment for what it's doing.

modmail/utils/plugins.py
```diff
@@ -0,0 +1,86 @@
...
+from modmail import plugins
+from modmail.log import ModmailLogger
+from modmail.utils.cogs import ExtMetadata
+from modmail.utils.extensions import BOT_MODE, unqualify
+
+log: ModmailLogger = logging.getLogger(__name__)
+
+
+BASE_PATH = Path(plugins.__file__).parent.resolve()
+PLUGIN_MODULE = "modmail.plugins"
```
bast0006, line 28: Thank you for moving this out of wherever I'd dumped it

modmail/utils/plugins.py
```diff
@@ -0,0 +1,86 @@
...
+PLUGINS: t.Dict[str, t.Tuple[bool, bool]] = dict()
+
+
+def walk_plugins() -> t.Iterator[t.Tuple[str, bool]]:
+    """Yield plugin names from the modmail.plugins subpackage."""
+    # walk all files in the plugins folder
+    # this is to ensure folder symlinks are supported,
+    # which are important for ease of development.
+    for path in glob.iglob(f"{BASE_PATH}/**/*.py", recursive=True):
```
bast0006, line 37: Note in the comment why we're not using Pathlib's glob--doesn't follow symlinks right. Even better, also link the python dev bug report I found for it..

bast0006: That'd be `https://bugs.python.org/issue29475`

bast0006 (changes requested):

modmail/extensions/extension_manager.py
bast0006: And everywhere else this is. It's in a lot of places. We should make this an atomic commit too, in case we go back and make ** mean something again.

onerandomusername:
Once @bast0006 re-reviews, I'm planning to merge this tomorrow.

modmail/__init__.py
```diff
@@ -26,7 +26,7 @@
 # file handler
 file_handler = logging.handlers.RotatingFileHandler(
     log_file,
-    maxBytes=5 * (2 ** 14),
+    maxBytes=5 * (2 ** 14),  # 81920 bytes, approximately 200 lines
```
bast0006: Alright. Now that we know what this is this is WAYYY too short. When I break a feature I generate more than 200 lines of output by accident, so I suggest we up this to 200k lines at least. That also lines up nicely with the megabyte barrier, which is where things go from easy upload to not easy upload, and certain tools start getting cumbersome. Just up this by `1000x` or 1024x if that's easier

modmail/extensions/extension_manager.py
```diff
@@ -198,7 +198,9 @@ async def rewalk_extensions(self, ctx: Context) -> None:
         log.debug(f"Refreshing list of {self.type}s.")
 
         # make sure the new walk contains all currently loaded extensions, so they can be unloaded
-        loaded_extensions = dict((en, sl) for en, sl in self.all_extensions.items() if en in self.bot.extensions)
+        loaded_extensions = dict(
+            (en, sl) for en, sl in self.all_extensions.items() if en in self.bot.extensions
+        )
```
bast0006: en -> ext_name, sl -> should_load

best to break this into a natural for loop, honestly.

bast0006 (dismissed):

modmail/__init__.py
onerandomusername: Done, upped it to 8MB

PR #46, opened 2021-08-07
chore: tone down codecov

onerandomusername:


.codecov.yml
```diff
@@ -10,20 +10,33 @@ coverage:
...
+        only_pulls: true
+    patch:
+      default:
+        # basic
+        target: auto
+        threshold: 5%
+        base: auto
+        # advanced
+        branches:
+          - master
```
bast0006, line 37: We don't have a master branch, do we?

.codecov.yml
```diff
@@ -10,20 +10,33 @@ coverage:
...
+        threshold: 5%
         base: auto
         flags:
           - unit
        # advanced settings
         branches:
           - main
+        if_ci_failed: ignore #success, failure, error, ignore
+        informational: true
+        only_pulls: true
```
bast0006, line 28: Also what does only pulls do here? Should that be commented or am I just ignorant on how this config works.

.codecov.yml
```diff
@@ -40,4 +53,4 @@ comment:
   require_changes: no
 
 github_checks:
-  annotations: true
+  annotations: false
```
bast0006, line 56: Praise.

bast0006 (changes requested):

.codecov.yml
onerandomusername: Its the available options, since the docs example has those comments, but also does not list them as possible options later.

onerandomusername: Sorry, only replied to the last part of your comment.

"only_pulls" makes codecov not run on commits that aren't part of a pull yet.

bast0006 (approved):

PR #48, opened 2021-08-07
ci: add python 3.10 to workflows

onerandomusername:
With python 3.10 set to release in a few months, it would be a good idea to ensure code works on it while developing it.

If it breaks at any point, that's a good sign of a regression.


onerandomusername:
Decided internally to wait until 3.10 is stable and our dependencies all officially support 3.10.

PR #56, opened 2021-08-19
Refactor startup

### LINKED ISSUE #57: chore[bot launcher]: refactor launch system
### Is there an existing issue for this?

- [X] I have searched the existing issues and there is no existing issue.

### Overview
Refactor launch system.

The current launch system involves a bunch of stuff in `__main__.py`, but that should all be in the `bot.run` bot method.

In addition, there should be a class method to create a bot with our custom configurations.

### Would you like to implement this yourself?

Yes, I'd like to implement this

onerandomusername:
- chore: [temp fix] add members intent
- [launch]: refactor bot creation and bot run
- chore: make status update on_ready

Closes #57

### DIFF
--- a/modmail/bot.py
+++ b/modmail/bot.py
@@ -4,6 +4,7 @@ import typing as t
 
 import arrow
 from aiohttp import ClientSession
+from discord import Intents
 from discord.ext import commands
 
 from modmail.config import CONFIG
@@ -26,7 +27,12 @@ class ModmailBot(commands.Bot):
         self.config = CONFIG
         self.http_session: t.Optional[ClientSession] = None
         self.start_time = arrow.utcnow()
-        super().__init__(command_prefix=commands.when_mentioned_or(self.config.bot.prefix), **kwargs)
+        intents = Intents(
+            guilds=True, messages=True, reactions=True, typing=True, members=True, emojis_and_stickers=True
+        )
+        super().__init__(
+            command_prefix=commands.when_mentioned_or(self.config.bot.prefix), intents=intents, **kwargs
+        )
 
     async def create_session(self) -> None:
         """Create an aiohttp client session."""
--- a/modmail/bot.py
+++ b/modmail/bot.py
@@ -71,6 +71,34 @@ class ModmailBot(commands.Bot):
             **kwargs,
         )
 
+    async def start(self, token: str, reconnect: bool = True) -> None:
+        """
+        Start the bot.
+
+        This function is called by the run method, and finishes the set up of the bot that needs an
+        asyncrhonous event loop running, before connecting the bot to discord.
+        """
+        try:
+            # create the aiohttp session
+            self.http_session = ClientSession(loop=self.loop)
+            self.logger.trace("Created ClientSession.")
+            # set start time to when we started the bot
+            self.start_time = arrow.utcnow()
+            # we want to load extensions before we log in, so that any issues in them are discovered
+            # before we connect to discord
+            self.load_extensions()
+            # next, we log in to discord, to ensure that we are able to connect to discord
+            await self.login(token)
+            # now that we're logged in and gotten a connection, we load all of the plugins
+            self.load_plugins()
+            # alert the user that we're done loading everything
+            self.logger.notice("Loaded all extensions, and plugins. Starting bot.")
+            # finally, we enter the main loop
+            await self.connect(reconnect=reconnect)
+        finally:
+            if not self.is_closed():
+                await self.close()
+
     def run(self, *args, **kwargs) -> None:
         """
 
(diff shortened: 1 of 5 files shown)


modmail/bot.py
```diff
@@ -19,27 +23,130 @@ class ModmailBot(commands.Bot):
...
+                loop.run_until_complete(start(*args, **kwargs))
+            except KeyboardInterrupt:
+                loop.run_until_complete(close())
+                # cancel all tasks lingering
+            finally:
+                loop.close()
+        .. warning::
+            This function must be the last function to call due to the fact that it
+            is blocking. That means that registration of events or anything being
+            called after this function call will not execute until it returns.
```
bast0006: This is wrong? start() is not an alternative for this method of startup, also, this isn't our language 👀

modmail/bot.py
```diff
@@ -19,27 +23,130 @@ class ModmailBot(commands.Bot):
...
-        await self.create_session()
-        await super().start(*args, **kwargs)
+        loop = self.loop
+
+        try:
+            loop.add_signal_handler(signal.SIGINT, lambda: loop.stop())
+            # this one we may want to get rid of, depending on certain things, and just hard stop instead.
+            loop.add_signal_handler(signal.SIGTERM, lambda: loop.stop())
+        except NotImplementedError:
+            pass
```
bast0006, line 91: This can be commented better: "Add signal handlers so the loop stops safely"

modmail/bot.py
```diff
@@ -124,3 +231,4 @@ def remove_cog(self, cog: str) -> None:
     async def on_ready(self) -> None:
         """Send basic login success message."""
         self.logger.info("Logged in as %s", self.user)
+        await self.change_presence(status=discord.Status.online, activity=self.activity)
```
bast0006, line 219: Don't put change_presence in on_ready, as on_ready may be fired weirdly. Bot init takes a status argument

modmail/bot.py
```diff
@@ -19,27 +23,130 @@ class ModmailBot(commands.Bot):
...
-    async def create_session(self) -> None:
-        """Create an aiohttp client session."""
-        self.http_session = ClientSession()
+    @classmethod
+    def create(cls, *args, **kwargs) -> "ModmailBot":
+        """
+        Create a ModmailBot instance.
 
-    async def start(self, *args, **kwargs) -> None:
+        This configures our instance with all of our custom options, without making __init__ unusable.
```
bast0006, line 41: We don't expect actual customization of the bot call. Providing the ability to is nice, but this is really just dodging a large `__init__` method and we're probably fine to just have a big one, especially when there's a lot of initialization, like here. Some of these should also be moved out of here, like intents

modmail/bot.py
```diff
@@ -19,27 +23,130 @@ class ModmailBot(commands.Bot):
...
+        """
+        intents = Intents(
+            guilds=True,
+            messages=True,
+            reactions=True,
+            typing=True,
+            members=True,
+            emojis_and_stickers=True,
+        )
```
bast0006, line 50: This should be driven by config somehow, or at least specified somewhere constant and not in here.

modmail/bot.py
```diff
@@ -19,27 +23,130 @@ class ModmailBot(commands.Bot):
...
+        # start with an invisible status while we load everything
+        status = discord.Status.invisible
+        activity = Activity(type=discord.ActivityType.listening, name="users dming me!")
+        # listen to messages mentioning the bot or matching the prefix
+        # ! NOTE: This needs to use the configuration system to get the prefix from the db once it exists.
+        prefix = commands.when_mentioned_or(CONFIG.bot.prefix)
+        description = "Modmail bot by discord-modmail."
+        # allow commands not caring of the case
+        case_insensitive_commands = True
```
bast0006, line 59: These lines are redundant

modmail/bot.py
```diff
@@ -19,27 +23,115 @@ class ModmailBot(commands.Bot):
...
+            future.remove_done_callback(stop_loop_on_completion)
+            self.logger.info("Cleaning up tasks.")
+            _cleanup_loop(loop)
+
+        if not future.cancelled():
+            try:
+                return future.result()
+            except KeyboardInterrupt:
+                # I am unsure why this gets raised here but suppress it anyway
+                return None
```
bast0006, line 134: Per discussion, this needs to be moved out of here so we don't end up God Objecting our way into misery.

bast0006 (changes requested):

modmail/bot.py
onerandomusername: our bot requires the member intent, so where would you suggest the intents go to?

modmail/bot.py
onerandomusername: These are the minimally required intents for our bot.

modmail/bot.py
onerandomusername: On_Ready

modmail/bot.py
bast0006: Globals

modmail/bot.py
```diff
@@ -19,27 +32,100 @@ class ModmailBot(commands.Bot):
...
+            self.http_session = ClientSession(loop=self.loop)
+            self.logger.trace("Created ClientSession.")
+            # set start time to when we started the bot
+            self.start_time = arrow.utcnow()
+            # we want to load extensions before we log in, so that any issues in them are discovered
+            # before we connect to discord
+            self.load_extensions()
+            # next, we log in to discord, to ensure that we are able to connect to discord
+            await self.login(token)
+            # now that we're logged in and gotten a connection, we load all of the plugins
```
bast0006, line 79: Nit: "have" a connection

tests/test_bot.py
```diff
@@ -27,21 +27,10 @@ def bot() -> ModmailBot:
 
     ModmailBot instance.
     """
-    bot: ModmailBot = ModmailBot()
+    bot: ModmailBot = ModmailBot(loop=asyncio.get_event_loop())
```
bast0006, line 30: Why are we directly passing the loop here? can't we just run get_event_loop from anywhere? So long as there's no threading.. but there should absolutely be _no threading_.

bast0006 (changes requested):

bast0006 (approved):

PR #58, opened 2021-08-19
chore[hosting]: use docker for hosting

onerandomusername:
Now that we have a dockerfile, one of our supported hosting platforms will use that file for running the bot, and will ignore the Procfile. This means that we can switch the process type back to worker, which is what our other supported hosting platform is expecting.


bast0006:
@onerandomusername have you tested this on our providers?

onerandomusername:
Yes, this resets it to what it was in commit ff75614c8eca3988a37431fd492f6a1712ed2666

bast0006:
Ok

bast0006 (approved):

onerandomusername:
We *could* add a [heroku.yml](https://devcenter.heroku.com/articles/build-docker-images-heroku-yml#heroku-yml-overview) file in order to build a docker container on heroku as well. Perhaps we should do that, which would hopefully let us remove requirements.txt in a future pr, provided every integration works after this is merged.

onerandomusername:
I've managed to set up a heroku.yml file, which now means that heroku will deploy by creating a docker image,
however, I had to run `heroku stack:set container` on the application, so I ended up making an app.json file.

This needs to be tested by deploying to heroku.

onerandomusername:
Testing if I can make this project deploy, and then we should be good to go.

dawnofmidnight (approved):
I cannot deploy to Heroku at the moment, but everything looks good.

PR #59, opened 2021-08-19
chore: make codecov support coverage for the entire modmail folder

onerandomusername:
since the line was missing a `/` codecov was only checking the modmail
file iteself, which doesn't exist, so no coverage was reported.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,7 +13,7 @@ coverage:
       default: false
       bot:
         paths:
-          - modmail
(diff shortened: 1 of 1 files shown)


onerandomusername:
There's no way to test that this worked w/out adding it to main, I think. 

Here's the docs if anyone wants to check them. https://docs.codecov.com/docs/commit-status

dawnofmidnight (approved):

PR #65, opened 2021-08-28
Feature: Database

### LINKED ISSUE #55: Main feature: database
We need a database module to store things persistently, such as thread state.

Things that need storing:
- Thread metadata (originating message, thread ID, user )
- Thread message data (who wrote, was moderator, highest role, message content, reactions? Attachments? )
- Config data (should just be a string/string KVS for now because that's how db config works)

These things should be server-keyed so in the future we can support multiserver if we want.

Shivansh-007:
Closes #55 

TODO...


onerandomusername:
Removed Shivansh as he's busy, @bast0006 is hopefully going to finish this pr.

PR #68, opened 2021-08-30
more tests

onerandomusername:
## Description

Continues expanding the test suite.  This attempts to cover *some* parts of the already installed portions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,6 @@
 import pytest
 
 
-def pytest_report_header(config):
+def pytest_report_header(config) -> str:
     """Pytest headers."""
     return "package: modmail"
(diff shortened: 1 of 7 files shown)


onerandomusername:
This pr can be reviewed at any time, once it is reviewed, other tests will be added as seperate prs.

onerandomusername:
> The new changes look good, the original code taken from pydis hasn't been reviewed.

The original code has been changed a somewhat substantial amount. The test_mocks file was rewritten to use pytest rather than unittest. While pytest supports unittests out of the box, I did not see an advantage to having unittests and pytest in the project.

On the mocks themselves, the context mock was updated to use the proper references. This now means that `ctx.channel.guild` and `ctx.guild` are the same mock, when they were previously different. In addition, a discord.Thread mock was added.

The last major change is the hash method will bitshift ids larger than 22.

PR #70, opened 2021-09-01
Use two lines after all imports instead of 1 line

onerandomusername:


Shivansh-007:
Changelog this.

onerandomusername:
Not worth a changelog, none of this is user facing.

bast0006:
Honestly, we should still changelog this, if only for ourselves

bast0006 (approved):
Alright... but this is gonna really shake things up a bit. Not sure how the final merge is going to be, but hopefully not too bad for all the other prs.

PR #71, opened 2021-09-01
Feat dispatcher

bast0006:
Add an event dispatcher for use with plugins and other inter-bot events.


modmail/dispatcher.py
```diff
@@ -0,0 +1,167 @@
...
+    blocking_handlers: Dict[str, List[Optional[CoroutineFunction]]]
+    blocking_priorities: Dict[str, List[int]]
+    handlers: Dict[str, List[CoroutineFunction]]
+
+    def __init__(self, *event_names: str):
+        self.handlers = {}
+        self.blocking_handlers = {}
+        self.blocking_priorities = {}
+
+        self.register_event(*event_names)
```
onerandomusername, line 30: Is it meant to not be possible to be able to make a Dispatcher without any events at all?

I feel like this line should not exist at all, and a Dispatcher must be made separately from registering events.

modmail/dispatcher.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def register_event(self, *event_names: str) -> None:
```
onerandomusername, line 32: Perhaps this could be more clear, as this can take multiple events, but the name of the function is singular.

modmail/dispatcher.py
```diff
@@ -0,0 +1,167 @@
...
+                args,
+                kwargs,
+            )
+            self.blocking_handlers[event_name] = []
+            self.handlers[event_name] = []
+            self.blocking_priorities[event_name] = []
+
+        for handler in self.blocking_handlers[event_name]:
+            if await handler(*args, **kwargs):
+                return
```
onerandomusername, line 165: Why does it return here with None? If its a blocking handler, it should only be blocked if the awaitable returns a Nonzero value.

This function should return the error code so the caller *knows* if the checks failed or not.

modmail/dispatcher.py
```diff
@@ -0,0 +1,167 @@
...
+from typing import Callable, Coroutine, Dict, List, Optional, Union
+
+from modmail import ModmailLogger
+
+logger: ModmailLogger = logging.getLogger(__name__)
+
+CoroutineFunction = Callable[..., Coroutine]
+
+
+class Dispatcher:
```
onerandomusername, line 13: Is there a way to close a dispatcher, to remove all events from it safely and ensure nothing was being awaited or uncollected?

tests/modmail/test_dispatcher.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def counter():
+        nonlocal tocks
+        tocks += 1
+        return tocks
+
+    return counter
+
+
+def make_mock_handler():
```
onerandomusername, line 24: Needs a Return type.

Actually, a lot of this section does.

modmail/dispatcher.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Dispatches events through an async event handler system.
+
+    Supports blocking events and priority.
+    """
+
+    # These are separate because it makes using the bisect module easier. See _register_handler.
+    blocking_handlers: Dict[str, List[Optional[CoroutineFunction]]]
+    blocking_priorities: Dict[str, List[int]]
```
onerandomusername, line 22: What are the supported priorities? How does this handle priorities between blocking_priorities and blocking_handlers. Some handlers may have a high priority, but aren't blocking.

onerandomusername (commented):
Is there not a bot attribute for this dispatcher? How is it used?

modmail/dispatcher.py
bast0006: I just added it for convenience. I'm not sure exactly how we intend to initialize all this, but if we want to init it with some events present, adding the line makes that possible

modmail/dispatcher.py
bast0006: We return none here in order to cancel further dispatch of the vent

modmail/dispatcher.py
bast0006: 👍

modmail/dispatcher.py
bast0006: That is something I should add.

bast0006: No, but that's because it's just a flow-through. Handling safe closure of all dispatch calls should ensure the dispatcher is cleaned up afterwards.

modmail/dispatcher.py
bast0006: Any integer priority level is supported. I suppose any float is also supported, but not explicitly. 

Handlers with a high priority but not "blocking" must necessarily be blocking as a group at minimum to prevent further events after them from being called too soon. However, because its much more complex to implement per-step blocking as opposed to just linear, I figured it's just much easier for everyone if they want a high priority handler to support the ability to block others. 

The only reason non-prioritized handlers don't support event cancellation is because they're all fired at once.

bast0006: I've added a nonblocking decorator to general utils if someone wants to explicitly opt into that behavior

onerandomusername (changes requested):
Found a major bug of not allowing methods of class instances to be passed self.

My test cog was like this:
```py
import logging

from discord.ext import commands
from discord.ext.commands import Context

from modmail.bot import ModmailBot
from modmail.dispatcher import Dispatcher
from modmail.log import ModmailLogger
from modmail.plugin_helpers import PluginCog

logger: ModmailLogger = logging.getLogger(__name__)

dispatch = Dispatcher()

dispatch.register_events("test")


class DispatchTest(PluginCog):
    """This is a DispatchTest."""

    def __init__(self, bot: ModmailBot):
        """INIT a TEST!."""
        self.bot = bot

    @dispatch.register("test")
    async def oof(self) -> None:
        """Test event."""
        print("hi 5")

    @commands.command()
    async def t(self, _: Context) -> None:
        """Test dispatch."""
        await dispatch.dispatch("test")


def setup(bot: ModmailBot) -> None:
    """Add the gateway logger to the bot."""
    bot.add_cog(DispatchTest(bot))
```

PR #73, opened 2021-09-06
fix: change flake8 instance to always run in the poetry venv in pre-commit

onerandomusername:
fixes problem of not being in the venv when running git commit


Shivansh-007 (approved):
Works well

PR #74, opened 2021-09-16
fix: install poetry with pipx when running in docker

onerandomusername:
Installs poetry into a venv with pipx for docker.

This means that our dependencies and poetry's dependencies will not conflict.


Dockerfile
```diff
@@ -1,22 +1,36 @@
 FROM python:3.9-slim
 
 # Set pip to have cleaner logs and no saved cache
+# Set poetry to not make virtualenvs
+# set pipx to use the same folder that pip puts scripts in
 ENV PIP_NO_CACHE_DIR=false \
-    POETRY_VIRTUALENVS_CREATE=false
+    POETRY_VIRTUALENVS_IN_PROJECT=true \
+    POETRY_INSTALLER_PARALLEL=false \
```
Shivansh-007, line 8: Whats the reason for adding this in? `POETRY_INSTALLER_PARALLEL=false`

Shivansh-007:
Could you also remove `.gitattributes`, we don't have that file anymore.

onerandomusername:
UPDATE: I never tested this before now, but poetry is not needed to install and use importlib.metadata

It is fully possible to use `pip install .` and importlib.metadata will still work.


I'll be redoing this pull request to best execute this.

PR #76, opened 2021-09-24
fix: paginator can now not use an embed

onerandomusername:
The paginator now will not always use an embed. This allows non-embedded pagination since embeds are restricted to a maximum width which we cannot control, but codeblocks are restrained to the width of the screen.

### DIFF
--- a/modmail/utils/pagination.py
+++ b/modmail/utils/pagination.py
@@ -62,7 +62,7 @@ class ButtonPaginator(ui.View, DpyPaginator):
         contents: Union[List[str], str],
         /,
         source_message: Optional[discord.Message] = None,
-        embed: Optional[Embed] = _AUTOGENERATE,
+        embed: Union[Embed, bool, None] = _AUTOGENERATE,
         timeout: float = 180,
         *,
         footer_text: str = None,
@@ -90,9 +90,11 @@ class ButtonPaginator(ui.View, DpyPaginator):
         self.suffix = suffix
         self.max_size = max_size
         self.linesep = linesep
-        if embed is _AUTOGENERATE:
+        if embed is _AUTOGENERATE or embed is True:
             self.embed = Embed()
         else:
+            if embed is False:
+                embed = None
             self.embed = embed
 
         # used if embed is None
--- a/modmail/utils/pagination.py
+++ b/modmail/utils/pagination.py
@@ -31,9 +31,12 @@ FORWARD_LABEL = "  \u276f  "  # >
 JUMP_LAST_LABEL = " \u276f\u276f "  # >>
 STOP_PAGINATE_EMOJI = "\u274c"  # [:x:] This is an emoji, which is treated differently from the above
 
-logger: ModmailLogger = logging.getLogger(__name__)
+NO_EMBED_FOOTER_BUMP = 15
 
 _AUTOGENERATE = object()
+
+
+logger: ModmailLogger = logging.getLogger(__name__)
 
 
 class ButtonPaginator(ui.View, DpyPaginator):
@@ -78,6 +81,8 @@ class ButtonPaginator(ui.View, DpyPaginator):
         If source message is provided and only_users is NOT provided, the paginator will respond
             to the author of the source message. To override this, pass an empty list to `only_users`.
 
+        By default, an embed is created. However, a custom embed can
+        be passed, or None can be passed to not use an embed.
         """
         self.index = 0
         self._pages: List[str] = []
@@ -92,11 +97,11 @@ class ButtonPaginator(ui.View, DpyPaginator):
 
         # used if embed is None
         self.content = ""
-        if embed is None:
+        if self.embed is None:
             self.title = title
             # need to set the max_size down a few to be able to set a "footer"
             # page indicator is "page xx of xx"
-            self.max_size -= 15
+            self.max_size -= NO_EMBED_FOOTER_BUMP + len(self.title or "")
             if self.title is not None:
                 self.max_size -= len(title)
             if footer_text is not None:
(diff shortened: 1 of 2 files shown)


modmail/utils/pagination.py
```diff
@@ -230,7 +252,9 @@ def update_states(self) -> None:
         if the paginator is on the last page, the jump last/move forward buttons will be disabled.
         """
         # update the footer
-        self.embed.set_footer(text=self.get_footer())
+        text = self.get_footer()
+        if self.embed:
+            self.embed.set_footer(text=text)
```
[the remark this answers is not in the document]
onerandomusername: get_footer does a few things other than just getting the footer, so it has to be called unconditionally.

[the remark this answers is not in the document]
bast0006: `get_footer` should then not do those things. I should have caught that. The name is currently misleading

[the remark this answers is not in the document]
onerandomusername: fixed in e54683d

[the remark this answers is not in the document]
onerandomusername: * 256011e49f292c6b9a75242d294bb956a0745f61

modmail/utils/pagination.py
```diff
@@ -251,9 +239,18 @@ def update_states(self) -> None:
...
+        else:
+            self.embed.description = self._pages[self.index]
+            page_indicator = f"Page {self.index+1}/{len(self._pages)}"
+            self.embed.set_footer(
+                text=(
+                    f"{self.footer_text} ({page_indicator})"
+                    if self.footer_text is not None
+                    else page_indicator
+                )
+            )
```
bast0006: This is awkward, can we un-one-line this? We can also not compare with None, as setting footer text to "" will result in not great output anyway, so we just check if it's present.
```py
if self.footer_text:
    footer_text = f"{self.footer_text} ({page_indicator})"
else:
    footer_text = page_indicator
```

bast0006 (changes requested):

modmail/utils/pagination.py
onerandomusername: How is this already outdated when no commits have been made? 🤔

onerandomusername: implemented in c9b33a3

modmail/utils/pagination.py
```diff
@@ -82,7 +85,22 @@ def __init__(
...
+        else:
+            self.embed = embed
+
+        # used if embed is None
+        self.content = ""
+        if embed is None:
+            self.title = title
+            # need to set the max_size down a few to be able to set a "footer"
+            # page indicator is "page xx of xx"
+            self.max_size -= 15
```
Shivansh-007, line 99: Make `15` a constant

modmail/utils/pagination.py
```diff
@@ -57,13 +59,14 @@ def __init__(
         contents: Union[List[str], str],
         /,
         source_message: Optional[discord.Message] = None,
-        embed: Embed = None,
+        embed: Optional[Embed] = _AUTOGENERATE,
```
Shivansh-007, line 62: Add a line about passing `None` for no embeds in the docstring.

modmail/utils/pagination.py
```diff
@@ -82,7 +85,22 @@ def __init__(
...
         self.linesep = linesep
-        self.embed = embed or Embed()
+        if embed is _AUTOGENERATE:
+            self.embed = Embed()
+        else:
+            self.embed = embed
+
+        # used if embed is None
+        self.content = ""
+        if embed is None:
```
Shivansh-007, line 95: self.embed

modmail/utils/pagination.py
onerandomusername: 170fa3a

modmail/utils/pagination.py
onerandomusername: 170fa3a

modmail/utils/pagination.py
onerandomusername: 170fa3a

Shivansh-007 (approved):
Not tested

onerandomusername:
Will rebase.

modmail/utils/pagination.py
```diff
@@ -75,14 +81,31 @@ def __init__(
...
         self._pages: List[str] = []
         self.prefix = prefix
         self.suffix = suffix
         self.max_size = max_size
         self.linesep = linesep
-        self.embed = embed or Embed()
+        if embed is _AUTOGENERATE:
+            self.embed = Embed()
+        else:
+            self.embed = embed
```
bast0006, line 96: Can we use `True` for autogenerate?

modmail/utils/pagination.py
```diff
@@ -212,15 +243,6 @@ async def interaction_check(self, interaction: Interaction) -> bool:
...
-    def get_footer(self) -> str:
-        """Returns the footer text."""
-        self.embed.description = self._pages[self.index]
-        page_indicator = f"Page {self.index+1}/{len(self._pages)}"
-        footer_txt = (
-            f"{self.footer_text} ({page_indicator})" if self.footer_text is not None else page_indicator
-        )
-        return footer_txt
-
     def update_states(self) -> None:
```
bast0006, line 246: Bikeshed: name this `update` to make more sense?

bast0006 (changes requested):

modmail/utils/pagination.py
onerandomusername: update states does not actually send the message, that is left to send_page

bast0006 (approved):

PR #75, opened 2021-09-19
feat/improve configuration

onerandomusername:
- feat: rewrite the entire configuration system to use marshmallow and attrs
- config: remove config_default.toml since its now part of the classes

Closes #22

PR #77, opened 2021-09-25
feat/error handler

onerandomusername:
- feat: add initial error handler
- add some CheckFailure handling support
- fix: implement remaining UserInputErrors
- fix: title regex
- fix: add concurrency errors and cooldown error support
- fix: don't log CommandNotFound errors unless in a dev mode
- fix: handle bot missing permissions

### DIFF
--- a/modmail/extensions/utils/error_handler.py
+++ b/modmail/extensions/utils/error_handler.py
@@ -8,7 +8,8 @@ from discord.ext import commands
 
 from modmail.bot import ModmailBot
 from modmail.log import ModmailLogger
-from modmail.utils.cogs import ExtMetadata, ModmailCog
+from modmail.utils.cogs import BotModes, ExtMetadata, ModmailCog
+from modmail.utils.extensions import BOT_MODE
 
 
 logger: ModmailLogger = logging.getLogger(__name__)
@@ -18,6 +19,8 @@ EXT_METADATA = ExtMetadata()
 ERROR_COLOUR = discord.Colour.red()
 
 ERROR_TITLE_REGEX = re.compile(r"(?<=[a-zA-Z])([A-Z])(?=[a-z])")
+
+ANY_DEV_MODE = BOT_MODE & (BotModes.DEVELOP.value + BotModes.PLUGIN_DEV.value)
 
 
 class ErrorHandler(ModmailCog, name="Error Handler"):
@@ -92,14 +95,14 @@ class ErrorHandler(ModmailCog, name="Error Handler"):
             logging.debug(f"Command {ctx.command} had its error already handled locally; ignoring.")
             return
 
-        if not isinstance(error, commands.CommandError):
-            logger.error("What in the world...")
-            return
-        logger.trace(error)
         if isinstance(error, commands.CommandNotFound):
             # ignore every time the user inputs a message that starts with our prefix but isn't a command
             # this will be modified in the future to support prefilled commands
+            if ANY_DEV_MODE:
+                logger.trace(error)
             return
+
+        logger.trace(error)
 
         title = None
         msg = None
--- a/modmail/extensions/utils/error_handler.py
+++ b/modmail/extensions/utils/error_handler.py
@@ -32,7 +32,7 @@ class ErrorHandler(ModmailCog, name="Error Handler"):
     @staticmethod
     def error_embed(message: str, title: str = None) -> discord.Embed:
         """Create an error embed with an error colour and reason and return it."""
-        return discord.Embed(message, colour=ERROR_COLOUR, title=title or "Error Occured")
+        return discord.Embed(message, colour=ERROR_COLOUR, title=title or "Error Occurred")
 
     @staticmethod
     def get_title_from_name(error: typing.Union[Exception, str]) -> str:
@@ -73,7 +73,7 @@ class ErrorHandler(ModmailCog, name="Error Handler"):
     async def handle_bot_missing_perms(
         self, ctx: commands.Context, error: commands.BotMissingPermissions
     ) -> bool:
-        """Handles bot missing perms by dming the user if they have a permission which may be to fix this."""
+        """Handles bot missing permissing by dming the user if they have a permission which may be able to fix this."""  # noqa: E501
         embed = self.error_embed(str(error))
         try:
             await ctx.send(embeds=[embed])
@@ -166,7 +166,7 @@ class ErrorHandler(ModmailCog, name="Error Handler"):
 
         elif isinstance(error, commands.CommandInvokeError):
             # generic error
-            logger.error(f'Error occured in command "{ctx.command}".', exc_info=error.original)
+            logger.error(f'Error occurred in command "{ctx.command}".', exc_info=error.original)
             # todo: this should log somewhere else since this is a bot bug.
             embed = self.error_embed(
                 "Oops! Something went wrong internally in the command you were trying to execute. "
--- a/modmail/extensions/utils/error_handler.py
+++ b/modmail/extensions/utils/error_handler.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import typing
 
 import discord
@@ -17,12 +18,23 @@ EXT_METADATA = ExtMetadata()
 ERROR_COLOUR = discord.Colour.red()
 
 
+def _raise(e: BaseException, *arg, **kwargs) -> typing.NoReturn:
+    raise e
+
+
 class ErrorHandler(ModmailCog, name="Error Handler"):
     """Handles all errors across the bot."""
 
     def __init__(self, bot: ModmailBot):
         self.bot = bot
 
+    @commands.check(lambda _: _raise(commands.NotOwner("oy")))
+    @commands.command()
+    async def error(self, ctx: commands.Context) -> None:
+        """."""
+        await ctx.send(5)
+
+    @commands.has_permissions(administrator=False)
     @commands.command()
     async def cmd(self, ctx: commands.Context, error: str = None, *args) -> None:
         """."""
@@ -41,22 +53,17 @@ class ErrorHandler(ModmailCog, name="Error Handler"):
     ) -> discord.Embed:
         """Handling deferred from main error handler to handle UserInputErrors."""
         embed = None
-        if isinstance(error, commands.BadArgument):
-            embed = self.error_embed(str(error), title="Bad argument")
-        elif isinstance(error, commands.MissingRequiredArgument):
-            embed = self.error_embed(str(error), title="Missing required argument")
-        elif isinstance(error, commands.TooManyArguments):
-            embed = self.error_embed(str(error), title="Too many arguments")
-        elif isinstance(error, commands.BadUnionArgument):
+        title = "User Input Error"
+        if isinstance(error, commands.BadUnionArgument):
             # TODO: complete
             msg = ""
-            embed = self.error_embed(msg, title="Bad union argument")
+            embed = self.error_embed(msg, title="Bad Union Argument")
         elif isinstance(error, commands.BadLiteralArgument):
             # TODO: complete
             msg = ""
-            embed = self.error_embed(msg, title="Bad literal argument")
+            embed = self.error_embed(msg, title="Bad Literal Argument")
         elif isinstance(error, commands.ArgumentParsingError):
-            embed = self.error_embed(str(error), title="Argument parsing error")
+            embed = self.error_embed(str(error), title="Argument Parsing Error")
             if isinstance(error, commands.UnexpectedQuoteError):
                 ...
             elif isinstance(error, commands.InvalidEndOfQuotedStringError):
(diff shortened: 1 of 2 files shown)


modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+from modmail.log import ModmailLogger
+from modmail.utils.cogs import BotModes, ExtMetadata, ModmailCog
+from modmail.utils.extensions import BOT_MODE
+
+
+logger: ModmailLogger = logging.getLogger(__name__)
+
+EXT_METADATA = ExtMetadata()
+
+ERROR_COLOUR = discord.Colour.red()
```
dawnofmidnight, line 19: Does this need to be a constant? It's just used once.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+class ErrorHandler(ModmailCog, name="Error Handler"):
+    """Handles all errors across the bot."""
+
+    def __init__(self, bot: ModmailBot):
+        self.bot = bot
+
+    @staticmethod
+    def error_embed(message: str, title: str = None) -> discord.Embed:
+        """Create an error embed with an error colour and reason and return it."""
+        return discord.Embed(message, colour=ERROR_COLOUR, title=title or "Error Occured")
```
dawnofmidnight, line 35: Occurred has two Rs

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            msg = self.get_title_from_name(str(error))
+            title = self.get_title_from_name(error)
+        else:
+            title = self.get_title_from_name(error)
+        return embed or self.error_embed(msg or str(error), title=title)
+
+    async def handle_bot_missing_perms(
+        self, ctx: commands.Context, error: commands.BotMissingPermissions
+    ) -> bool:
+        """Handles bot missing perms by dming the user if they have a permission which may be to fix this."""
```
dawnofmidnight, line 76: Also "may be to fix this" is a bit confused. Maybe it could be replaced with "could be used to fix this"? Or maybe just "may be able to fix this"?

dawnofmidnight (changes requested):
This looks pretty good, although I can't test at the moment because I'm not on my usual dev environment. One side question: should "dm" or "dming" be capitalized?

modmail/extensions/utils/error_handler.py
onerandomusername: Yes, in a future pr I'll be moving it to the configuration system.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+
+logger: ModmailLogger = logging.getLogger(__name__)
+
+EXT_METADATA = ExtMetadata()
+
+ERROR_COLOUR = discord.Colour.red()
+
+ERROR_TITLE_REGEX = re.compile(r"(?<=[a-zA-Z])([A-Z])(?=[a-z])")
+
+ANY_DEV_MODE = BOT_MODE & (BotModes.DEVELOP.value + BotModes.PLUGIN_DEV.value)
```
bast0006, line 23: Is there not a better way to do this from ExtMetadata?

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+class ErrorHandler(ModmailCog, name="Error Handler"):
+    """Handles all errors across the bot."""
+
+    def __init__(self, bot: ModmailBot):
+        self.bot = bot
+
+    @staticmethod
+    def error_embed(message: str, title: str = None) -> discord.Embed:
+        """Create an error embed with an error colour and reason and return it."""
+        return discord.Embed(message, colour=ERROR_COLOUR, title=title or "Error Occurred")
```
bast0006, line 35: `title` should not be nullable here. Let's enforce some kind of automated naming, like derived from the exception name. Plus, it's not typehinted as Optional.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            self._reset_command_cooldown(ctx)
+        embed = None
+        msg = None
+        title = "User Input Error"
+        if isinstance(error, commands.BadUnionArgument):
+            msg = self.get_title_from_name(str(error))
+            title = self.get_title_from_name(error)
+        else:
+            title = self.get_title_from_name(error)
+        return embed or self.error_embed(msg or str(error), title=title)
```
bast0006, line 71: This can be cleaned up a little. Try moving `str(error)` into the `else` clause, and then you don't need the or. Also no need for `title` here to be a keyword argument.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+
+    async def handle_bot_missing_perms(
+        self, ctx: commands.Context, error: commands.BotMissingPermissions
+    ) -> bool:
+        """Handles bot missing permissing by dming the user if they have a permission which may be able to fix this."""  # noqa: E501
+        embed = self.error_embed(str(error))
```
bast0006, line 77: In the interest of removing "Error Ocurred", we could either hardcode in "Permissions Failure"/"Missing Permissions" or something here, to help localize embed naming to what's appropriate. I'd like to avoid any chance of "Oops Something Went Wrong," as those kinds of errors are _extremely_ frustrating to get, and repetitive embeds are large and obnoxious.

modmail/extensions/utils/error_handler.py
bast0006: Perhaps ExtMetadata or BotModes should have an `any_dev_mode` property?

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+                if perms.value & ctx.channel.permissions_for(ctx.author).value:
+                    logger.info(
+                        f"Attempting to dm {ctx.author} since they have a permission which may be able "
+                        "to give the bot send message permissions."
+                    )
+                    try:
+                        await ctx.author.send(embeds=[embed])
+                    except discord.Forbidden:
+                        logger.notice("Also encountered an error when trying to reply in dms.")
+                        return False
```
bast0006, line 99: I love this feature. Needs to be behind a configuration toggle, though.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            title = "DMs Only"
+        elif isinstance(error, commands.NoPrivateMessage):
+            title = "Server Only"
+        elif isinstance(error, commands.BotMissingPermissions):
+            # defer handling BotMissingPermissions to a method, since this can be problematic
+            await self.handle_bot_missing_perms(ctx, error)
+            return None
+        else:
+            title = self.get_title_from_name(error)
+        embed = self.error_embed(msg or str(error), title=title)
```
bast0006, line 121: msg here is always str(error), so we can simplify that out

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+        msg = None
+        if isinstance(error, commands.CheckAnyFailure):
+            title = self.get_title_from_name(error.checks[-1])
+            msg = str(error)
+        elif isinstance(error, commands.PrivateMessageOnly):
+            title = "DMs Only"
+        elif isinstance(error, commands.NoPrivateMessage):
+            title = "Server Only"
+        elif isinstance(error, commands.BotMissingPermissions):
+            # defer handling BotMissingPermissions to a method, since this can be problematic
```
bast0006, line 116: Add to this comment? Instead of just "Since this can be problematic" how about "handled specially for better UX and detail" or similar? We're doing good things, not just handling bad things, so that should be apparent without having to go digging into another method. It'll also direct someone looking for how the Missing Permissions DM works to the right function.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            # ignore every time the user inputs a message that starts with our prefix but isn't a command
+            # this will be modified in the future to support prefilled commands
+            if ANY_DEV_MODE:
+                logger.trace(error)
+            return
+
+        logger.trace(error)
+
+        title = None
+        msg = None
```
bast0006, line 141: These aren't used

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+        # TODO: this has a fundamental problem with any BotMissingPermissions error
+        # if the issue is the bot does not have permissions to send embeds or send messages...
+        # yeah, problematic.
+
+        if not should_respond:
+            logger.debug("Not responding to error since should_respond is falsey.")
+            return
+
+        if embed is None:
+            embed = self.error_embed(msg or str(error), title=title or self.get_title_from_name(error))
```
bast0006, line 188: Since neither title nor  message are actually used from the above, they can be dropped. This should also be commented, I think:
`# general case` or something like that.

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            if isinstance(error.original, discord.Forbidden):
+                await self.handle_bot_missing_perms(ctx, error.original)
+                return
+
+        # TODO: this has a fundamental problem with any BotMissingPermissions error
+        # if the issue is the bot does not have permissions to send embeds or send messages...
+        # yeah, problematic.
+
+        if not should_respond:
+            logger.debug("Not responding to error since should_respond is falsey.")
```
bast0006, line 184: Try perhaps "Since the embed has already been sent or belongs to a hidden command and thus should be hidden."

modmail/extensions/utils/error_handler.py
```diff
@@ -0,0 +1,195 @@
...
+        elif isinstance(error, commands.CheckFailure):
+            embed = await self.handle_check_failure(ctx, error)
+            # handle_check_failure may send its own error if its a BotMissingPermissions error.
+            if embed is None:
+                should_respond = False
+        elif isinstance(error, commands.ConversionError):
+            # s = object()
+            # error.converter.convert.__annotations__.get("return", s)
+            # embed = error
+            ...
```
bast0006, line 156: Incomplete? @onerandomusername
If we actually want incomplete fallthrough, a 
`pass  # fall through to default response formatting` would be the best way to annotate it I think.

bast0006 (changes requested):

modmail/extensions/utils/error_handler.py
onerandomusername: Configuration system is in a different pr, GH-75, so I'll add a constant for now.

modmail/extensions/utils/error_handler.py
onerandomusername: I don't want to add that to the metadata system, also metadata got a bit of a refactor in GH-69.

What I could add to the cog system is this constant, and just compute it there instead.

onerandomusername:
There is a fundamental issue with the botmissingpermissions error handling. In almost all send message cases, the bot will make a second case which will 100% fail.

The bot should check its cache *or* request the channel perms to see if it has send message perms. If it does, should also check for embed image permissions, and send a standard message of content if that fails. 

if *that* is not a valid permission, the bot should see if it has add_reaction permissions and maybe add a `:x:` if it can.

modmail/extensions/utils/error_handler.py
onerandomusername: Yeah, I think I meant to do something but can remove this.

modmail/extensions/utils/error_handler.py
onerandomusername: Actually, this was behind a partial config toggle. If the bot is in a developer mode it will the dm the user, but was not turned on for main users.

modmail/extensions/utils/error_handler.py
onerandomusername: There is a place where its already dervied from the exception class. That exists elsewhere, and is used for most names. I suppose I could remove the default here, since its only used in one place and call the method with that default. 

However, if making it required, I'm switching the argument order as well.

modmail/extensions/utils/error_handler.py
onerandomusername: I'll switch this from the developer default to the configuration system once it is merged.

modmail/extensions/utils/error_handler.py
```diff
@@ -60,54 +60,64 @@ async def handle_user_input_error(
...
-                # non-general permissions
-                perms = discord.Permissions(
-                    administrator=True,
-                    manage_threads=True,
-                    manage_roles=True,
-                    manage_channels=True,
+            responded = True
+        elif bot_perms >= discord.Permissions(send_messages=True):
+            # make a message as similar to the embed, using as few permissions as possible
+            # this is the only place we send a standard message instead of an embed, so no helper methods
```
bast0006, line 81: .. are neccessary

modmail/extensions/utils/error_handler.py
```diff
@@ -60,54 +60,64 @@ async def handle_user_input_error(
...
+            )
+            # intentionally not setting responded to True, since we want to attempt to dm the user
+            logger.warning(
+                f"Missing partial required permissions for {ctx.channel}. "
+                "I am able to send messages, but not embeds."
+            )
+        else:
+            logger.error(f"Unable to send an error message to channel {ctx.channel}")
+
+        if responded is not True and ANY_DEV_MODE:
```
bast0006, line 94: not responded is clearer

modmail/extensions/utils/error_handler.py
```diff
@@ -162,30 +167,36 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
                 msg = f"Command `{ctx.invoked_with}` is disabled."
                 if reason := ctx.command.extras.get("disabled_reason", None):
                     msg += f"\nReason: {reason}"
-                embed = self.error_embed(msg, title="Command Disabled")
+                embed = self.error_embed("Command Disabled", msg or str(error))
```
bast0006: msg is always present here, no need for the or

modmail/extensions/utils/error_handler.py
```diff
@@ -162,30 +167,36 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
...
+                should_respond = False
+            else:
+                # generic error
+                logger.error(f'Error occurred in command "{ctx.command}".', exc_info=error.original)
+                # todo: this should log somewhere else since this is a bot bug.
+                embed = self.error_embed(
+                    "Error Occurred",
+                    "Oops! Something went wrong internally in the command you were trying to execute. "
+                    "Please report this error and what you were trying to do to the developers.",
+                )
```
bast0006, line 185: Is this always going to be a bot bug, or will plugins fire this too?

onerandomusername:
> 
> 
> There is a fundamental issue with the botmissingpermissions error handling. In almost all send message cases, the bot will make a second case which will 100% fail.
> 
> The bot should check its cache _or_ request the channel perms to see if it has send message perms. If it does, should also check for embed image permissions, and send a standard message of content if that fails.
> 
> if _that_ is not a valid permission, the bot should see if it has add_reaction permissions and maybe add a `:x:` if it can.

This is implemented in 7f19996

modmail/extensions/utils/error_handler.py
onerandomusername: Plugins and internal features will fire this, so I've customised it in 34e7696 to detect the name of the module it was fired from.

bast0006 (approved):

PR #78, opened 2021-09-27
enhancement: embedify all of the meta commands

onerandomusername:
Old responses:
image

New responses:
image


Shivansh-007 (approved):
Yes

onerandomusername:
This pr should wait until GH-75 is merged as it will cause merge conflicts, and I would rather resolve them here than on that pr.

onerandomusername:
Actually, the conflicts are easy enough to fix there, so I'll merge this now.

PR #80, opened 2021-10-04
maybe fix workflows not running

onerandomusername:


onerandomusername:
Figured this out, branches must be up to date with main since that requirement is turned on. This must be a way that github is trying to save on action usage or something....

PR #89, opened 2021-10-17
Custom change-logger

Shivansh-007:
Closes #87 

#### To Do (Now):

- [x] Move constants to `config.toml`
- [x] Add GitHub workflow for checking news
- [x] Document the code
- [x] Clean and remove redundant code

#### To Do (Will be done once the current API/design is finalised and reviewed):

- [x] Add documentation
- [ ] Port existing change-log

PR #111, opened 2021-11-08
scripts: add wrapper to run scripts by invoking the scripts module

onerandomusername:
### Description
Because there are so many upcoming scripts, (GH-75, GH-89) this adds a way to be able to shorten the interface to launch a script.

Rather than use `python -m scripts.export_requirements`, `python -m scripts export_reqs` can be used. Internally, scripts.export_requirements thinks it has been called the same way as if it was called directly.

This script can be thought of as a launcher for the scripts within the scripts module.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -193,7 +193,7 @@ unicode_backport = ["unicodedata2"]
 
 [[package]]
 name = "click"
-version = "8.0.1"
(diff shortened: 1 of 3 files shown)


Shivansh-007 (approved):
Add it as a task, otherwise LGTM!

PR #116, opened 2021-11-15
hooks: export docs/.requirements.txt too

onerandomusername:


scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
     return hash == get_hash(content)
 
 
-def main(req_path: os.PathLike, should_validate_hash: bool = True) -> typing.Optional[int]:
+def _extract_packages_from_requirements(requirements: str) -> list:
```
Shivansh-007, line 60:
```suggestion
def _extract_packages_from_requirements(requirements: str) -> set[str]:
```

scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
...
+
+    # parse the packages out of the requirements txt
+    packages = _extract_packages_from_requirements(contents)
+
+    # get the version of each package
+    packages_metadata: dict = toml["package"]
+    new_versions = {}
+    for package in packages_metadata:
+        if package["name"].lower().replace("_", "-") in packages:
+            new_versions[package["name"]] = package["version"]
```
Shivansh-007, line 126:
```suggestion
    new_versions = {
        package["name"]: package["version"]
        for package in packages_metadata
        if package["name"].lower().replace("_", "-") in packages
    }
```

scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
...
+        return 0
+    with open(file, "w") as f:
+        f.write(new_contents)
+
+    return 1
+
+
+def main(
+    req_path: os.PathLike, should_validate_hash: bool = True, export_doc_requirements: bool = True
+) -> typing.Optional[int]:
```
Shivansh-007, line 139:
```suggestion
) -> int:
```

scripts/export_requirements.py
```diff
@@ -149,16 +230,23 @@ def main(req_path: os.PathLike, should_validate_hash: bool = True) -> typing.Opt
...
+        code = _export_doc_requirements(lockfile, DOC_REQUIREMENTS)
+    else:
+        code = 0
+
     if req_path.exists():
         with open(req_path, "r") as f:
             if req_txt == f.read():
                 # nothing to edit
-                return 0
+                return 0 or code
```
Shivansh-007, line 243: `0` is falsy so wouldn't it always return `code`?

scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
...
+
+def _export_doc_requirements(toml: dict, file: pathlib.Path, *packages) -> int:
+    """
+    Export the provided packages versions.
+
+    Return values:
+    0 no changes
+    1 exported new requirements
+    2 file does not exist
+    3 invalid packages
```
Shivansh-007, line 108: It doesn't return `3` anywhere?

scripts/export_requirements.py
```diff
@@ -19,10 +19,12 @@
 
 
 GENERATED_FILE = pathlib.Path("requirements.txt")
+DOC_REQUIREMENTS = pathlib.Path("docs/.requirements.txt")
 
 VERSION_RESTRICTER_REGEX = re.compile(r"(?P<sign>[<>=!]{1,2})(?P<version>\d+\.\d+?)(?P<patch>\.\d+?|\.\*)?")
 PLATFORM_MARKERS_REGEX = re.compile(r'sys_platform\s?==\s?"(?P<platform>\w+)"')
 
+PACKAGE_REGEX = re.compile(r"^(?P<package>[^=<>~]+)")
```
Shivansh-007, line 27: This can be changed to just match `==` as we never edit the file manually and it only gets edited by this script, where it only uses `==`.

scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
...
+            continue
+        if not len(line.strip()):
+            continue
+
+        # requirement files we will be parsing can have `;` or =<>~
+        match = PACKAGE_REGEX.match(line)
+        if match is None:
+            continue
+
+        package = match.group("package").strip().lower().replace("_", "-")
```
Shivansh-007, line 75: Why do we do `.replace("_", "-")`? I don't see any packages using `_`.

scripts/export_requirements.py
```diff
@@ -55,7 +57,86 @@ def get_hash(content: dict) -> str:
...
+            continue
+        if not len(line.strip()):
+            continue
+
+        # requirement files we will be parsing can have `;` or =<>~
+        match = PACKAGE_REGEX.match(line)
+
+        package = match.group("package").strip()
+        try:
+            req[i] = package + "==" + packages[package.lower().replace("_", "-")]
```
Shivansh-007, line 94: Wouldn't the package already be lower cased and have no `_` as we do these changes on line 76.

scripts/export_requirements.py
Shivansh-007: This comment can be ignored, see: https://discord.com/channels/798235512208490526/845824657306288158/909740574322610177

Shivansh-007 (changes requested):

scripts/export_requirements.py
onerandomusername: yes, but I left it returning `0 or` here so it would be clear that this part returns zero, so it is deferred to code

scripts/export_requirements.py
onerandomusername: These are different functions.

scripts/export_requirements.py
Shivansh-007: > would be clear that this part returns zero


Huh? `0` is falsy so it would never be returned (it would always return `code`) unless `code` itself is `0`.

scripts/export_requirements.py
onerandomusername: Pypi replaces all `-` with `_` (or vice versa, they're the same character) and while poetry is supposed to normalize them to the lockfile, I don't want to come across a weird poetry bug that doesn't

scripts/export_requirements.py
Shivansh-007: Ah yes, I thought packages is coming from that function...which isn't the case.

But when we form this `packages` variable, can we just take the key as the edited name (on line 125), this would also reduce the redundancy to run the edit twice.

scripts/export_requirements.py
Shivansh-007: Hmm, I think it would be better to let poetry handle it and in case we come across such a bug in future we could add this edit functionality again.

In case, you still want that to stay, you should document why you do that in the code.

scripts/export_requirements.py
onerandomusername: Made the above function return a result which is used on this function.

scripts/export_requirements.py
onerandomusername: It does now, good eye

scripts/export_requirements.py
onerandomusername: Updated and made a comment about it

scripts/export_requirements.py
onerandomusername: Fixed

scripts/export_requirements.py
onerandomusername: Neat

scripts/export_requirements.py
onerandomusername: Updated, thanks

scripts/export_requirements.py
```diff
@@ -55,7 +56,96 @@ def get_hash(content: dict) -> str:
...
+
+        packages.add(package)
+
+    return packages, req
+
+
+def _update_versions_in_requirements(requirements: "list[str]", packages: dict) -> str:
+    """Update the versions in requirements with the provided package to version mapping."""
+    req = requirements
+    for i, package in enumerate(req.copy()):
```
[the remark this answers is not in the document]
onerandomusername: we modify req in this for loop.
its not safe to modify it while iterating through it

scripts/export_requirements.py
```diff
@@ -55,7 +56,96 @@ def get_hash(content: dict) -> str:
...
+        package = req[i].lower().replace("_", "-")
+
+        packages.add(package)
+
+    return packages, req
+
+
+def _update_versions_in_requirements(requirements: "list[str]", packages: dict) -> str:
+    """Update the versions in requirements with the provided package to version mapping."""
+    req = requirements
```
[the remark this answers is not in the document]
onerandomusername: this was just because i wanted the parameter to stay requirements... uhhh I'll fix that.

Shivansh-007 (approved):

PR #118, opened 2021-11-16
logging: add config for trace loggers and logging directory

onerandomusername:
- allow configuring loggers and their specified levels

Some loggers can get spammy when debugging, this allows
turning them down, or setting other loggers to lower levels than the root

To use, set a MODMAIL_{level}_LOGGERS environment variable, delimiting
the loggers with `,`
Valid levels are all of the valid levels for logging, as follows
  trace, debug, info, notice, warning, error, critical

- add support for configuring the file logging directory

The directory for logging files was fully dependent on the current
working directory
This caused my environment to be littered with logging directories

The solution to this was to continue to use the current working
directory, unless the parent directory of the bot is also a parent of
the cwd, in which case the bot parent directory is used for creating
the logging directory.

In addition, `MODMAIL_LOGGING_DIRECTORY` has been added as an override
environment variable. Setting it will use that directory for logging.


modmail/log.py
```diff
@@ -1,7 +1,105 @@
...
+
+def set_logger_levels() -> None:
+    """
+    Set all loggers to the provided environment variables.
+
+    eg MODMAIL_TRACE_LOGGERS will be split by `,` and each logger will be set to the trace level
+    This is applied for every logging level.
+    """
+    env_vars = _get_env()
+    fmt_key = "MODMAIL_{level}_LOGGERS"
```
onerandomusername, line 64: TBH I'm not sure if this should be this format, or if `MODMAIL_LOGGERS_{level}` makes more sense since the variable is now more hierarchical.

onerandomusername: I'm going to reverse this format, actually.

PR #120, opened 2021-11-17
Tests: patch os.environ to exclude tests

onerandomusername:
Tests should have a repeatable environment, so this solves that problem by patching os.environ and clearing during testing.


.gitignore
```diff
@@ -104,6 +104,7 @@ celerybeat.pid
 *.sage.py
 
 # Environments
+!test.env
```
Shivansh-007, line 107: Make the location for this more specific, I sometimes use `test.env` for storing secrets while I am testing something.

Shivansh-007 (approved):

tests/modmail/test.env
```diff
@@ -0,0 +1 @@
+MODMAIL_BOT_TOKEN='NjQzOTQ1MjY0ODY4MDk4MDQ5.342b.4inDLBILY69LOLfyi6jk420dpyjoEVsCoModM'
```
onerandomusername, line 1: For the record, this is a fake token.

PR #122, opened 2021-11-19
deps: upgrade to 3.10 supported wheels

onerandomusername:
Upgrade packages that don't have wheels on python 3.10 to packages that... do have wheels.

NOTE: aiohttp still requires a wheel to be built since discord.py is blocking it,
and discord.py doesn't get updates anymore.

cchardet also still requires being built.

This doesn't mean that we won't support python 3.10, but that building on 3.10 can take a slight bit longer time.

Full diff:
```diff
13c13
< cffi==1.14.6; python_version >= "3.6"
---
> cffi==1.15.0; python_version >= "3.6"
55c55
< multidict==5.1.0; python_version >= "3.6" and python_full_version >= "3.8.0"
---
> multidict==5.2.0; python_version >= "3.6" and python_full_version >= "3.8.0"
67c67
< pycares==4.0.0; python_version >= "3.6"
---
> pycares==4.1.2; python_version >= "3.6"
103c103
< yarl==1.6.3; python_version >= "3.6" and python_full_version >= "3.8.0"
---
> yarl==1.7.2; python_version >= "3.6" and python_full_version >= "3.8.0"
```


Shivansh-007 (approved):

onerandomusername:
Skipping changelog since this does *not* change what `pip install .` will install, which installs whatever resolved packages.

PR #123, opened 2021-11-19
ci: checkout the actual ref instead of github's merge commit

onerandomusername:
Github makes a merge commit for any pull requests.
This checks back out our custom branch for linting and testing.
This does means that we lose the benefits of being able to see what
would happen if we merged with our default branch, but the problems
of our custom pre-commit hooks getting different output, and different
installed dependencies from what is on the branch dependending on how
up-to-date our branches are, is a problem that I would rather not have.


onerandomusername:
Looked a little more into this and while it does fix one issue, it causes other issues with github and actions. Not sure the best way to fix this except to just make a branch up to date with the main branch when issues occur like this...

PR #124, opened 2021-11-19
feat: debugger module for dumping logs in guild

onerandomusername:
Add a small debugger cog for debugging bot issues.

Adds a dump command for logs (which doesn't work right now...) and a log command to paginate the last 100 lines of logs.

This should be useful when having to troubleshoot problems, as a user won't need to access the console for every error.


Shivansh-007:
I don't see the need for this.

onerandomusername:
This is aimed more for consumers, to help troubleshoot when something goes wrong.

Shivansh-007:
Viewing those logs in your terminal/editor is much more helpful than viewing them on discord... so I don't see how that helps.

PR #126, opened 2021-11-21
tests: fix environment not being replaced soon enough

onerandomusername:
continuation of #120

That solution using a fixture did not patch the environment soon enough. This will patch the environment and save a cache of the original environment, so it can replaced upon the end of a test run.


onerandomusername:
After further inspection, this should keep the environment unmodified, only removing all variables prefixed with MODMAIL_ in order to keep the environment more realistic of what would happen on a real environment, and only controlling the program specific environment variables.

onerandomusername:
However, environment variables are currently not prefixed with `MODMAIL_`, as that is introduced in #75.

onerandomusername:
This can be reviewed and merged, but will only have an effect on the testing environment once GH-75 is merged

onerandomusername:
Merging main to get the updated workflows here.

Will rebase before merge.

tests/modmail/conftest.py
```diff
@@ -6,6 +6,11 @@
 import pytest
 
 
+_ORIG_ENVIRON = None
+
+_modmail_env_prefix = "MODMAIL_"
```
Shivansh-007, line 11: This use `modmail.config.ENV_PREFIX` once config PR is merged, worth adding a `TODO`

tests/modmail/conftest.py
```diff
@@ -23,17 +28,35 @@ def _get_env():
     return pathlib.Path(__file__).parent / "test.env"
 
 
+def _get_env_vars() -> dict:
+    result = {}
+    for key, value in os.environ.items():
+        # not using upper() here is not a bug since the config system is case sensitive
```
Shivansh-007, line 34:
```suggestion
        # not using upper() since the config system is case sensitive
```

PR #133, opened 2021-11-27
Add lock no update task

Shivansh-007:
Please approve, kthx


onerandomusername (approved):
lgtm.

The reason to alphabetize is because taskipy will show them in the order they're defined here....

onerandomusername:
A little slow there, coveralls.

PR #138, opened 2024-09-25
update lint config

onerandomusername:
Update CI so its functional again 🙃


onerandomusername:
lint is as much as workable. tests will be a separate PR

